Issue #2, opened 2016-09-13
ExpressionBuilder should be assignable to StatementBuilder


yjbanov:
An expression is also a statement. For example:

```dart
void main() {
  "blah".substring(1);  // expression used as statement
}
```

So this should be allowed:

```dart
var methodBuilder = new MethodBuilder(...);
methodBuilder.addStatement(new ExpressionBuilder.invoke('substring'd));
```

matanlurey:
Sorry, let me clarify, you'd like to do:

```dart
var method = new MethodBuilder.returnVoid('main');
var statement = new StatementBuilder()
  ..addExpression(const LiteralString('blah').invoke('substring', positional: [
    const LiteralInt(1)
  ]));
method.addStatement(statement);
```

Is that right? Or do you _specifically_ mean that maybe we should allow:

```dart
// Option 1
new StatementBuilder.fromExpression(...)

// Option 2
expression.toStatement()
```

yjbanov:
The end result would be like in the first example, but the way I was thinking about is that there wouldn't be any special API to add an expression. Instead, an expression is a special case of a statement. IOW, `ExpressionBuilder extends StatementBuilder` so you can pass an `ExpressionBuilder` to `addExpression` or anything else that takes a `StatementBuilder`.

In Dart you can use any expression as a statement, e.g.:

```dart
// A valid Dart program
void main() {
  5;
  null;
  'hello';
  []..add(4);
}
```

yjbanov:
Or, maybe we should do what [the analyzer is doing](https://www.dartdocs.org/documentation/analyzer/0.29.0-alpha.0/analyzer/ExpressionStatement-class.html) :)

yjbanov:
In that case, option 2 looks great.

PR #4, opened 2016-09-14
Add Expression#toStatement, and tests.

matanlurey:
Closes https://github.com/dart-lang/code_builder/issues/2.

### DIFF
--- a/lib/src/expression_builder.dart
+++ b/lib/src/expression_builder.dart
@@ -5,11 +5,24 @@
 part of code_builder;
 
 // Shared functionality between ExpressionBuilder and _LiteralExpression.
-final Token _closeP = new Token(TokenType.CLOSE_PAREN, 0);
-final Token _openP = new Token(TokenType.OPEN_PAREN, 0);
-final Token _semicolon = new Token(TokenType.SEMICOLON, 0);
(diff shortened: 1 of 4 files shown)


lib/src/expression_builder.dart
```diff
@@ -59,18 +85,74 @@ abstract class ExpressionBuilder implements CodeBuilder<Expression> {
...
+    Map<String, CodeBuilder<Expression>> named: const {},
+  }); // TODO(matanl): Rename to invoke when factory is removed.
+
   /// Returns wrapped as a [ExpressionFunctionBody] AST.
   ExpressionFunctionBody toFunctionBody() => _asFunctionBody(this);
 
   /// Returns wrapped as a [FunctionExpression] AST.
   FunctionExpression toFunctionExpression() => _asFunctionExpression(this);
+
+  /// Returns as a [StatementBuilder].
```
yjbanov: nit: for doc comment, how about "Converts to" rather that "Returns as" to underscore that it's not a simple cast but a conversion.

yjbanov: Also, an example might be helpful:

For example, converts expression `null` to statement `null;`.

yjbanov:
LGTM

PR #9, opened 2016-09-15
Add a simple identifier Scope class and a test.

matanlurey:


lib/src/scope.dart
```diff
@@ -0,0 +1,65 @@
+part of code_builder;
+
+/// Determines an [Identifier] given where it appears and where it is from.
```
yjbanov: Did you mean "depending on where it appears..."?

lib/src/scope.dart
```diff
@@ -0,0 +1,65 @@
...
+  /// Create a default scope context.
+  ///
+  /// Actual implementation is _not_ guaranteed, only that all import prefixes
+  /// will be unique in a given scope (actual implementation may be naive).
+  factory Scope() = _IncrementingScope;
+
+  /// Create a context that does _not_ apply any scoping.
+  factory Scope.identity() = _IdentityScope;
+
+  /// Given a [symbol] and it's known [importUri], return an [Identifier].
```
yjbanov: nit: s/it's/its/

lib/src/scope.dart
matanlurey: Yup! Done.

lib/src/scope.dart
matanlurey: 👍

PR #10, opened 2016-09-16
Add support for library scoping

matanlurey:


lib/src/builders/parameter_builder.dart
```diff
@@ -9,74 +9,77 @@ part of code_builder;
...
 /// class-member (field) variable (see the `field` property in the
 /// constructors).
-class ParameterBuilder extends _AbstractCodeBuilder<FormalParameter> {
+class ParameterBuilder extends ScopeAware<FormalParameter> {
   static final Token _this = new KeywordToken(Keyword.THIS, 0);
 
+  final String _name;
+  final bool _isField;
+  final bool _isOptional;
+  final bool _isNamed;
```
yjbanov: Is this not the same as `_name.isNotEmpty`?

matanlurey: Whether it's a _named_ optional parameter. I should probably just make an enum, but it's not like I'm going to be adding more anytime soon.

Issue #26, opened 2016-11-16
Need a ClosureBuilder


alorenzen:
Right now, there's no way to create a closure expression. All we have is lamba() for MethodBuilders.

Example:
args.map((String arg) => doSomething(arg));

PR #29, opened 2016-11-17
Add additional features for NG2 output AST parity

matanlurey:
Closes https://github.com/dart-lang/code_builder/issues/26
Closes https://github.com/dart-lang/code_builder/issues/27

## 1.0.0-alpha+3

- BREAKING CHANGE: Added generics support to `TypeBuilder`:

`importFrom` becomes a _named_, not positional argument, and the named
argument `genericTypes` is added (`Iterable<TypeBuilder>`).

```dart
// List<String>
new TypeBuilder('List', genericTypes: [reference('String')])
```

- Added generic support to `ReferenceBuilder`:

```dart
// List<String>
reference('List').toTyped([reference('String')])
```

- Fixed a bug where `ReferenceBuilder.buildAst` was not implemented
- Added `and` and `or` methods to `ExpressionBuilder`:

```dart
// true || false
literal(true).or(literal(false));

// true && false
literal(true).and(literal(false));
```

- Added support for creating closures - `MethodBuilder.closure`:

```dart
// () => true
new MethodBuilder.closure(
  returns: literal(true),
  returnType: lib$core.bool,
)
```

### DIFF
--- a/lib/src/builders/method.dart
+++ b/lib/src/builders/method.dart
@@ -9,6 +9,7 @@ import 'package:code_builder/code_builder.dart';
 import 'package:code_builder/dart/core.dart';
 import 'package:code_builder/src/builders/annotation.dart';
 import 'package:code_builder/src/builders/class.dart';
+import 'package:code_builder/src/builders/expression.dart';
 import 'package:code_builder/src/builders/parameter.dart';
 import 'package:code_builder/src/builders/shared.dart';
 import 'package:code_builder/src/builders/statement.dart';
@@ -165,10 +166,18 @@ abstract class ConstructorBuilder
 
 /// Lazily builds a method/function AST when the builder is invoked.
 abstract class MethodBuilder
-    implements HasAnnotations, HasParameters, HasStatements, ValidClassMember {
+    implements
+        ExpressionBuilder,
+        HasAnnotations,
+        HasParameters,
+        HasStatements,
+        ValidClassMember {
   /// Creates a new [MethodBuilder].
-  factory MethodBuilder(String name,
-      {ExpressionBuilder returns, TypeBuilder returnType}) {
+  factory MethodBuilder(
+    String name, {
+    ExpressionBuilder returns,
+    TypeBuilder returnType,
+  }) {
     if (returns != null) {
       return new _LambdaMethodBuilder(
         name,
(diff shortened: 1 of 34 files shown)


lib/src/builders/method.dart
```diff
@@ -347,7 +395,27 @@ class _MethodBuilderImpl extends Object
         _property = property;
 
   @override
-  AstNode buildAst([Scope scope]) => buildFunction(scope);
+  AstNode buildAst([Scope scope]) {
+    if (_name != null) {
```
alorenzen: More of a personal style, but I like to use => and ternary operator here: 
buildAst() => _name != null ? buildFunction() : buildExpression();

lib/src/builders/method.dart
```diff
@@ -347,7 +395,27 @@ class _MethodBuilderImpl extends Object
...
+      return buildFunction(scope);
+    }
+    return buildExpression(scope);
+  }
+
+  @override
+  Expression buildExpression([Scope scope]) {
+    return new FunctionExpression(
```
alorenzen: Why not just use => ?

lib/src/builders/reference.dart
```diff
@@ -52,6 +59,36 @@ class ReferenceBuilder extends Object
...
+
+class _TypedReferenceBuilder extends ReferenceBuilder {
+  final List<TypeBuilder> _genericTypes;
+
+  _TypedReferenceBuilder(
+    this._genericTypes,
+    String name,
+    String importFrom,
+  )
+      : super._(name, importFrom);
```
alorenzen: Indentation seems off here. Dartfmt?

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: code_builder
-version: 1.0.0-alpha+2
+version: 1.0.0-alpha+3
```
alorenzen: I'm not really sure about the standard practice here, but should we bump the version # given a breaking change? Or because this is all still alpha, we're not too worried?

lib/src/builders/method.dart
matanlurey: Same as above

lib/src/builders/method.dart
matanlurey: Honestly just because dartfmt doesn't do a great job with > 80 char lambdas.

I can change it in a future PR

lib/src/builders/reference.dart
matanlurey: This is dartfmt :-/

lib/src/builders/method.dart
alorenzen: Fair enough. Like I said, it's just my personal preference.

pubspec.yaml
alorenzen: SGTM

PR #36, opened 2016-11-29
Begin taking advantage of analyzer's new AstFactory class.

stereotype441:
The old mechanism for creating analyzer ASTs (factory constructors on
the interface classes) will be going away in analyzer 0.30.


stereotype441:
@matanlurey @alorenzen

PR #55, opened 2017-01-05
More test cases, useful until we have more docs

matanlurey:
/cc @zoechi


zoechi:
@matanlurey thanks a lot! I had quite a hard time getting started, but it's getting better and I have a bit of working code already where I can look things up. The tests were helpful as well.

PR #67, opened 2017-01-11
Update class.dart to remove an unused import

matanlurey:


natebosch:
Any idea why this didn't show up in Travis? Are we running analyzer tests?

matanlurey:
@natebosch No idea: https://github.com/dart-lang/code_builder/blob/master/tool/presubmit.sh#L16

Looks like it should have been. I know the analyzer fails internally more strictly than externally, maybe it's a matter of changing what is considered a warning or error.

PR #68, opened 2017-01-17
rename MethodModifier constant names

yjbanov:
`async` identifier name is forbidden within `async` context. For example, the following is illegal Dart code:

```dart
test('illegal', () async {
  MethodModifier.async;
});
```


matanlurey:
@yjbanov WDUT about `asAsync` and `asAsyncStar` to be consistent with other methods like `asStatic`?

yjbanov:
Aye! I'll fix tomorrow.

yjbanov:
Done

PR #102, opened 2017-03-30
Run dartfmt to fix the build.

matanlurey:


matanlurey:
@natebosch I think dartfmt changed :-/

zoechi:
The change looks more like a dartfmt regression to me than an improvment

PR #108, opened 2017-05-02
Update analyzer

danrubel:
This change updates the internals to use the newer analyzer API so that code_builder can be used with both analyzer 0.29.11 and 0.30.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.0.2
+
+* update internals to use newer analyzer API
+
 ## 1.0.1
 
 * Support the latest version of `package:dart_style`.
--- a/lib/src/builders/method.dart
+++ b/lib/src/builders/method.dart
@@ -35,19 +35,19 @@ ConstructorBuilder constructorNamed(
 /// Various types of modifiers for methods.
 class MethodModifier implements ValidMethodMember {
   static const MethodModifier asAsync = const MethodModifier._(
-    'async',
+    Keyword.ASYNC,
     false,
   );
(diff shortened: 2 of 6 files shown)


matanlurey:
I'm OOO until Thursday so maybe Nate or Kevin or Andrew can help?

danrubel:
Travis is unhappy because `dartfmt` passes on stable but fails for older 1.22.

kevmoo:
Looks like you need to `dartfmt` @danrubel

alorenzen:
Kevin, I think there may be a breaking change to dartfmt between versions. We can get either 1.22 or 1.23 passing, but not both.

danrubel:
@kevmoo 
* `dartfmt` was failing on stable for unrelated file (file that I did not change) `test/e2e_test.dart`
* I [formatted that file](https://github.com/dart-lang/code_builder/pull/108/commits/941afb16b6a27735a6adf0b23d22ab9d736ff9c3) as part of this PR (see above)
* Now `dartfmt` on stable is passing but failing on 1.22 (see comment above)

Not sure how to make two different versions of `dartfmt` happy about a file that I do not need to change.
How should I proceed?

kevmoo:
We just need to ignore the lint for the other run

danrubel:
Is there a variable I can check in [tool/presubmit.sh](https://github.com/dart-lang/code_builder/blob/master/tool/presubmit.sh) to only perform the `dartfmt` check on stable?

kevmoo:
You can look at ENV

danrubel:
Presubmit updated. PTAL

PR #114, opened 2017-06-25
Started on V2, not using ASTs...

matanlurey:
... as requested by the Dart analyzer/front end team.

Started using `built_value` (since it gives it a nicer API anyway) and modeled method bodies as code blocks instead, since in reality it makes writing code generation that much easier if you only use typed structures for the API.


matanlurey:
Oops, wrong branch.

PR #115, opened 2017-06-25
Started on V2, not using ASTs...

matanlurey:
... as requested by the Dart analyzer/front end team.

Started using built_value (since it gives it a nicer API anyway) and modeled method bodies as code blocks instead, since in reality it makes writing code generation that much easier if you only use typed structures for the API.

[I realize this change is scary, hence why it's going on the `V2` branch instead until complete]


lib/src/emitter.dart
```diff
@@ -0,0 +1,322 @@
...
+    visitCode(spec.code, output);
+    output.write(' ');
+    return output;
+  }
+
+  @override
+  visitClass(Class spec, [StringSink output]) {
+    output ??= new StringBuffer();
+    spec.docs.forEach(output.writeln);
+    spec.annotations.forEach((a) => visitAnnotation(a, output));
```
alorenzen: Avoid using forEach - https://www.dartlang.org/guides/language/effective-dart/usage#avoid-using-iterableforeach-with-a-function-literal

lib/src/emitter.dart
```diff
@@ -0,0 +1,322 @@
...
+        ..writeAll(spec.mixins.map<StringSink>(visitType), ',');
+    }
+    if (spec.implements.isNotEmpty) {
+      output
+        ..write(' implements ')
+        ..writeAll(spec.implements.map<StringSink>(visitType), ',');
+    }
+    output.write(' {');
+    spec.constructors.forEach((c) => visitConstructor(c, spec.name, output));
+    spec.fields.forEach((f) => visitField(f, output));
```
alorenzen: Same here, just use a for-loop.

lib/src/visitors.dart
```diff
@@ -0,0 +1,79 @@
...
+  T visitSpec(Spec spec) => spec.accept(this);
+
+  @override
+  T visitType(TypeReference spec) => null;
+
+  @override
+  T visitTypeParameters(Iterable<TypeReference> specs) => null;
+}
+
+class RecursiveSpecVisitor<T> extends SimpleSpecVisitor<T> {
```
alorenzen: TODO here for implementation?

matanlurey:
@alorenzen Yeah, I think I can re-add scoping and imports in the next PR, just wanted to make sure I was covering all feature requirements for the injection package first so I can land this internally.

lib/src/emitter.dart
matanlurey: Will fix style nits in next PR.

lib/src/emitter.dart
matanlurey: Will fix style nits in next PR.

lib/src/visitors.dart
matanlurey: Will fix or remove in next PR.

PR #116, opened 2017-07-08
More work towards feature parity of 1x

matanlurey:
Added:

* `Allocator` (prefixing)
* `Directive` and `File`


matanlurey:
Ping @alorenzen since you're back - this is probably enough to roll I imagine.

lib/src/emitter.dart
```diff
@@ -2,18 +2,30 @@
...
+import 'specs/file.dart';
 import 'specs/method.dart';
 import 'specs/reference.dart';
 import 'specs/type_reference.dart';
 import 'visitors.dart';
 
-class DartEmitter extends GeneralizingSpecVisitor<StringSink> {
-  const DartEmitter();
+class DartEmitter implements SpecVisitor<StringSink> {
+  final Allocator _allocator;
```
alorenzen: I don't think _allocator.imports is ever called. So how would you actually print out the imports in a file?

test/specs/file_test.dart
```diff
@@ -0,0 +1,29 @@
...
+import 'package:code_builder/code_builder.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('File', () {
+    const $linkedHashMap = const Reference('LinkedHashMap', 'dart:collection');
+    test('should emit a source file', () {
+      expect(
+        new File((b) => b
+          ..directives.add(new Directive.import('dart:collection'))
```
alorenzen: Is the assumption that you always have to add the imports manually? Can you add a test which uses allocator.imports instead?

Also, maybe a test case for prefixing imports?

lib/src/allocator.dart
```diff
@@ -0,0 +1,95 @@
...
+  /// style and instead takes a conservative approach of prefixing _every_
+  /// import except references to `dart:core` (which are considered always
+  /// imported).
+  factory Allocator.simplePrefixing() = _PrefixedAllocator;
+
+  /// Returns a reference string given a [reference] object.
+  ///
+  /// For example, a no-op implementation:
+  /// ```dart
+  /// allocate(const Reference('List', 'dart:core')); // Outputs 'List'.
```
matanlurey: Done.

lib/src/allocator.dart
```diff
@@ -0,0 +1,95 @@
...
+  /// ```
+  ///
+  /// Where-as an implementation that prefixes imports might output:
+  /// ```dart
+  /// allocate(const Reference('Foo', 'package:foo')); // Outputs '_i1.Foo'.
```
matanlurey: Done.

lib/src/emitter.dart
matanlurey: Thanks, good catch. Added to `visitFile` and updated `file_test`.

test/specs/file_test.dart
matanlurey: Fixed. Added test cases.

PR #131, opened 2017-10-10
More code support across codebase

matanlurey:
This is ~probably enough to implement anything else (statements/expressions).


lib/src/specs/annotation.dart
```diff
@@ -24,7 +24,11 @@ abstract class Annotation
   Code get code;
 
   @override
-  R accept<R>(SpecVisitor<R> visitor) => visitor.visitAnnotation(this);
+  R accept<R>(
+    SpecVisitor<R> visitor, [
+    R context,
```
alorenzen: In my opinion, when you only have the two parameters, putting each on it's own line doesn't add any value, and wastes quite a bit of vertical space.

Remove the trailing , so that dartfmt puts this all on one line.

This will also keep it consistent with the formatting in lib/src/specs/code/expression.dart

Same for the other instances of this method.

matanlurey: Ack, I have another change upcoming, I'll make it there.

PR #132, opened 2017-10-11
Even more expression/code support

matanlurey:
Closing the gap on the `1.x` releases.

Notably, still missing basic things like arguments to invocations, but next PR will add them.


lib/src/specs/expression.dart
```diff
@@ -0,0 +1,136 @@
...
+  @override
+  visitLiteralExpression(LiteralExpression expression, [StringSink output]) {
+    output ??= new StringBuffer();
+    return output..write(expression.literal);
+  }
+
+  @override
+  visitLiteralListExpression(
+    LiteralListExpression expression, [
+    StringSink output,
```
alorenzen: Drop the trailing , to be consistent with the rest of the file.

matanlurey: Will address in follow-up PR.

PR #133, opened 2017-10-12
Boom expression args

matanlurey:
Requires https://github.com/dart-lang/code_builder/pull/132.

This adds arguments to invocations, `literalString`, `literalMap`, and a bit of refactors.


matanlurey:
Ping @alorenzen 😉

lib/src/specs/expression.dart
```diff
@@ -0,0 +1,229 @@
...
+  T visitLiteralListExpression(LiteralListExpression expression, [T context]);
+  T visitLiteralMapExpression(LiteralMapExpression expression, [T context]);
+}
+
+/// Knowledge of how to write valid Dart code from [ExpressionVisitor].
+///
+/// **INTERNAL ONLY**.
+abstract class ExpressionEmitter implements ExpressionVisitor<StringSink> {
+  @override
+  visitBinaryExpression(BinaryExpression expression, [StringSink output]) {
```
alorenzen: Return values?

matanlurey: It's automatically inherited in strong-mode semantics (hence why travis is green).

I don't mind explicitly adding, but doesn't add any obvious value in the VM.

Issue #138, opened 2017-10-13
2.0: Add Expression.return


matanlurey:

matanlurey:
+ `Expression.awaited`

PR #140, opened 2017-10-14
Fix a number of issues.

matanlurey:
Closes https://github.com/dart-lang/code_builder/issues/135.
Closes https://github.com/dart-lang/code_builder/issues/136.
Closes https://github.com/dart-lang/code_builder/issues/137.
Closes https://github.com/dart-lang/code_builder/issues/138.
Closes https://github.com/dart-lang/code_builder/issues/139.

~~This API already exists internally, so I'm skipping tests until we are back in sync - I won't release this as `2.0.0-alpha+3` before adding test coverage, though.~~ **EDIT**: Added tests anyway.

### DIFF
--- a/lib/src/specs/expression.dart
+++ b/lib/src/specs/expression.dart
@@ -67,9 +67,15 @@ abstract class Expression implements Spec {
   }
 
   /// Return `var {name} = {this}`.
-  Expression assignVar(String name) {
+  Expression assignVar(String name, [Reference type]) {
     return new BinaryExpression._(
-      new LiteralExpression._('var $name'),
+      type == null
+          ? new LiteralExpression._('var $name')
+          : new BinaryExpression._(
+              type.toExpression(),
+              new LiteralExpression._(name),
+              '',
+            ),
       this,
       '=',
     );
@@ -86,7 +92,7 @@ abstract class Expression implements Spec {
               '',
             ),
       this,
-      '=',
+      '$name =',
     );
   }
 
@@ -101,7 +107,7 @@ abstract class Expression implements Spec {
               '',
             ),
       this,
-      '=',
+      '$name =',
     );
   }
 
--- a/lib/src/specs/expression.dart
+++ b/lib/src/specs/expression.dart
@@ -39,15 +39,83 @@ abstract class Expression implements Spec {
     return new BinaryExpression._(toExpression(), other, '&&');
   }
 
+  /// This expression preceding by `await`.
+  Expression get awaited {
+    return new BinaryExpression._(
+      const LiteralExpression._('await'),
+      this,
+      '',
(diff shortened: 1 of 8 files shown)


lib/src/specs/expression.dart
```diff
@@ -39,15 +39,89 @@ abstract class Expression implements Spec {
     return new BinaryExpression._(toExpression(), other, '&&');
   }
 
+  /// This expression preceding by `await`.
```
yjbanov: nit: "preceded"

lib/src/specs/expression.dart
```diff
@@ -39,15 +39,89 @@ abstract class Expression implements Spec {
...
+  Expression get awaited {
+    return new BinaryExpression._(
+      const LiteralExpression._('await'),
+      this,
+      '',
+    );
+  }
+
+  /// Return `{other} = {this}`.
+  Expression assign(Expression other) {
```
yjbanov: Hmm, `a.assign(b)` reads way too much like `a = b`. Maybe flip the logic and make it `{this} = {other}`?

Or perhaps rename to `assignTo`? Although not sure that helps much.

lib/src/specs/expression.dart
```diff
@@ -39,15 +39,89 @@ abstract class Expression implements Spec {
...
+    return new BinaryExpression._(
+      other,
+      this,
+      '=',
+    );
+  }
+
+  /// Return `{other} ??= {this}`.
```
yjbanov: Same issue as the one with `assign`.

lib/src/specs/expression.dart
```diff
@@ -65,27 +139,74 @@ abstract class Expression implements Spec {
...
       this,
       positionalArguments,
       namedArguments,
+      typeArguments,
+      null,
     );
   }
 
-  /// May be overriden to support other types implementing [Expression].
+  /// Returns a const instance of this expression with a named constructor.
```
yjbanov: I'm not sure what this does. Could you provide an example (and add it to the dartdoc)? Specifically, I'm not aware of many kinds of expressions that you can _invoke_ with a named constructor, i.e. only references to classes. Perhaps this method should be moved to `Reference`?

lib/src/specs/expression.dart
```diff
@@ -65,27 +139,74 @@ abstract class Expression implements Spec {
...
+    return new InvokeExpression._const(
+      this,
+      positionalArguments,
+      namedArguments,
+      typeArguments,
+      name,
+    );
+  }
+
+  /// This expression preceding by `return`.
```
yjbanov: nit: `preceded`

matanlurey:
Going to fix some of the comments in the next CL.

PR #156, opened 2017-11-03
Escape single quotes in String literals

### LINKED ISSUE #155: Escape quotes in stringLiteral expressions
A String with a `'` will not work. Since it's not clear whether double or single quotes are used we might want to escape quotes.

natebosch:
Closes #155

If there is an unescaped literal in a String the code would otherwise be
invalid.
If some code is doing the escaping before creating the literal this will
end up double-escaping.

### DIFF
--- a/lib/src/specs/expression/literal.dart
+++ b/lib/src/specs/expression/literal.dart
@@ -48,9 +48,16 @@ Expression literalNum(num value) => new LiteralExpression._('$value');
 ///
 /// **NOTE**: The string is always formatted `'<value>'`.
 ///
-/// If [raw] is `true`, creates a raw String formatted `r'<value>'`.
-Expression literalString(String value, {bool raw: false}) =>
-    new LiteralExpression._("${raw ? 'r' : ''}'$value'");
+/// If [raw] is `true`, creates a raw String formatted `r'<value>'` and the
+/// value may not contain a single quote.
+/// If [raw] is `false` escapes single quotes in the value.
+Expression literalString(String value, {bool raw: false}) {
+  if (raw && value.contains('\'')) {
+    throw new ArgumentError('Cannot include a single quote in a raw string');
+  }
+  final escaped = value.replaceAll('\'', '\\\'');
+  return new LiteralExpression._("${raw ? 'r' : ''}'$escaped'");
+}
 
 /// Creates a literal list expression from [values].
 LiteralListExpression literalList(List<Object> values, [Reference type]) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 2.1.0
 
 * Added support for raw `String` literals.
+* Automatically escapes single quotes in now-raw `String` literals.
 
 ## 2.0.0
 
--- a/test/specs/code/expression_test.dart
+++ b/test/specs/code/expression_test.dart
@@ -18,6 +18,14 @@ void main() {
     expect(literalString(r'$monkey', raw: true), equalsDart(r"r'$monkey'"));
   });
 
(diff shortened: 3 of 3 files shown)


lib/src/specs/expression/literal.dart
```diff
@@ -48,9 +48,16 @@ Expression literalNum(num value) => new LiteralExpression._('$value');
...
-/// If [raw] is `true`, creates a raw String formatted `r'<value>'`.
-Expression literalString(String value, {bool raw: false}) =>
-    new LiteralExpression._("${raw ? 'r' : ''}'$value'");
+/// If [raw] is `true`, creates a raw String formatted `r'<value>'` and the
+/// value may not contain a single quote.
+/// If [raw] is `false` escapes single quotes in the value.
+Expression literalString(String value, {bool raw: false}) {
+  if (raw && value.contains('\'')) {
+    throw new ArgumentError('Cannot include a single quote in a raw string');
+  }
```
lrhn: Consider moving this escaping to the emitting point. There is nothing inherently wrong with a string literal containing `'`, but you have to be careful when you try to emit it as source code. That's the best position to do any escaping, not at creation when you don't know how the string will be used.

Instead of disallowing `'` in a string, you can switch to double- or triple-quoted string literals when needed:
```dart
  var x = r"There's nothing wrong with apostophes!";
  var y = r'''There's nothing wrong with apostophes!''';
```
(I *prefer* double quoted strings for readable text, partly because of this - and partly because I prefer them for all strings :smile:)

lib/src/specs/expression/literal.dart
```diff
@@ -48,9 +48,16 @@ Expression literalNum(num value) => new LiteralExpression._('$value');
...
+  final escaped = value.replaceAll('\'', '\\\'');
```
lrhn: You might want to check whether the caller has already escaped the apostrophe.
Otherwise:
```dart
  var text = literalString(r"I\'m calling the ops!");
```
will generate an invalid invalid code anyway.

lib/src/specs/expression/literal.dart
natebosch: This is the point at which we determine what code will get output - we could maybe delay that more to the point where the emitter visits us but it would take some refactoring and I'm not sure we'd end up making any different decisions.

I do like the idea of being smart enough to change our quote type based on the presence of values we'd need to escape. I also like the idea of interpreting the value like you describe.

@matanlurey - can you help me understand the motivation behind allowing the caller to specify `raw`? Is this critical or could we add some magic to make that decision for them as described?

matanlurey: @kevmoo needed this!

kevmoo: > @kevmoo needed this!

Because I didn't want to do *this* hard work...

natebosch: I don't think it would be terribly difficult to get the semantics described - "a string literal that evaluates to `value`" and I do think it would improve usability.

There are two tricky bits:
1. This probably means eliminating the `raw:` argument. @matanlurey - do you think this is worth a breaking change?
2. Handling the case where we *want* String interpolation.

`literalString(r"$foo")` - if we want a String in source that evaluates to this value we could choose `'\$foo'` or `r'$foo'`.
How do we get this String in the source though: `'$foo'`?

We could potentially introduce a new concept... `interpolatedString(Iterable<Expression>)`
If I want in source `'Hello ${person.name} my name is $greeter'` it would look like: `interpolatedString([literalString('Hello '), refer('person').property('name'), literalString(' my name is ', refer('greeter')])`

Very clunky but maybe necessary? I'm also unsure whether we could actually make this API work...

PR #158, opened 2017-11-03
Only support 2.0.0-dev.

matanlurey:
Due to function types I think we need this anyway.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -5,7 +5,11 @@ author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/code_builder
 
 environment:
-  sdk: '>=1.24.0 <2.0.0'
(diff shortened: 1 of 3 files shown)


pubspec.yaml
```diff
@@ -19,3 +23,7 @@ dev_dependencies:
   built_value_generator: '>=2.0.0 <5.0.0'
   source_gen: '^0.7.0'
   test: ^0.12.0
+
+transformers:
+  - test/pub_serve:
+      $include: test/**_test.dart
```
natebosch: [nit] AFAIK the `**` here isn't doing anything that `*` wouldn't.

matanlurey: Recursive? I'll just keep it for now.

natebosch: I think recursive only applies when its is followed by a `/`

PR #160, opened 2017-11-10
Add '_i' to import prefixes

natebosch:
This matches old behavior and is a little more clear than just a digit.
This will also satisfy the lint on the latest SDK. See
https://github.com/dart-lang/linter/issues/684#issuecomment-343543273

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.1.1-dev
+
+* Imports are prefixed with `_i1` rather than `_1` which satisfies the lint
+  `lowercase_with_underscores`.
+
 ## 2.1.0
 
 We now require the Dart 2.0-dev branch SDK (`>= 2.0.0-dev`).
--- a/README.md
+++ b/README.md
@@ -83,9 +83,9 @@ void main() {
 
 Outputs:
 ```dart
-import 'package:a/a.dart' as _1;
-import 'package:b/b.dart' as _2;
+import 'package:a/a.dart' as _i1;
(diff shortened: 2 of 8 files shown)


natebosch:
ping @matanlurey - is this worth getting in?

matanlurey:
Sorry I missed this.

Would you consider this a breaking change? Should it be opt-in for now?

Otherwise many golden file tests will fail, for example.

@natebosch

natebosch:
Yeah I have some golden tests that I expect to start failing after this change.

I personally wouldn't consider that breaking - I think its a test that is (perhaps necessarily) relying on an implementation detail and so I'd expect it to be brittle.

I'd rather not make it an option. The API to enable it becomes another part of the public surface area which is broken if/when we remove the option, and we deal with higher complexity in the mean time.

I don't have strong opinions here. I know we've had customers who want these generated files to follow that lint rule, but I'd be fine with either of:
1 - push it and break brittle tests
2 - hold off until 3.0 and ship it then

matanlurey:
It doesn't bother me either way. 

We _should_ document that the Allocator _doesn't_ guarantee and output as part of this PR then?

@natebosch

natebosch:
Added a note in the Doc comment for `Allocator.simplePrefixing`

matanlurey (approved):

matanlurey:
@natebosch Let me know if you want a release this week.

natebosch:
No rush - this change was a "may as well do it while it's on my mind" rather than a "I need this right now".

PR #162, opened 2017-11-20
Accept an Iterable in `literalList`

natebosch:
Since the LiteralListExpression is in the public API we shouldn't change
it's field type, so immediately call .toList()


natebosch:
on our next breaking version bump we should consider removing `LiteralListExpression` from the public API. I *think* the only thing the user code should need to see this as is an `Expression`.

Thoughts?

matanlurey:
One reason for keeping it public is technically you can roll your own Emitter.

natebosch:
Ah, that makes sense.

PR #167, opened 2017-11-22
Add EqualsDart and EqualsDart.format.

matanlurey:
The rationale here is to remove a dependency on `dart_style` and `analyzer` in `3.0.0`.

In reality it is only depended on for `equalsDart`, which is only for testing. If in the future we get modular packages or sub packages then it might be worth bringing back in some form, but this likely is to give more flexibility, especially around packages like `analyzer` that are almost impossible to revision against.

I've made the changes to make the tests in _this_ package future-proof.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -14,6 +14,7 @@ web:
 dependencies:
   built_collection: ^1.0.0
   built_value: '>=2.0.0 <5.0.0'
+  # TODO: Remove in 3.0.0.
   dart_style: ^1.0.0
   matcher: ^0.12.0
   meta: ^1.0.5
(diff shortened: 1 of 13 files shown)


pubspec.yaml
```diff
@@ -14,13 +14,15 @@ web:
...
   built_value: '>=2.0.0 <5.0.0'
+  # TODO: Remove in 3.0.0.
   dart_style: ^1.0.0
   matcher: ^0.12.0
   meta: ^1.0.5
 
 dev_dependencies:
   build_runner: '>=0.4.0 <0.6.0'
   built_value_generator: '>=2.0.0 <5.0.0'
+  dart_style: ^1.0.0
```
natebosch: I thought pub would yell if the same dependency shows up twice?

matanlurey: You're right. Removed the dupe.

matanlurey:
@natebosch: `matcher` as a dependency isn't too bad:

https://github.com/dart-lang/matcher/blob/master/pubspec.yaml

... just `stack_trace`. For some reason.

PR #168, opened 2017-11-23
Deprecate the Annotation class.

matanlurey:
Largely was just boilerplate over some string-based expression 95%+ of the time.

This change shouldn't be breaking, but will allow us to simply remove `Annotation` in 3.x.

### DIFF
--- a/lib/src/specs/annotation.dart
+++ b/lib/src/specs/annotation.dart
@@ -31,6 +31,7 @@ abstract class Annotation extends Expression
   /// This property exists in order to keep compatibility with having a concrete
   /// [Annotation] class, but also allow migrating to using [Expression] instead
   /// to add annotations.
+  @protected
   Code get codeForAnnotation;
 
   @override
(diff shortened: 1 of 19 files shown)


lib/src/specs/annotation.dart
```diff
@@ -10,29 +10,45 @@ import 'package:meta/meta.dart';
...
   Annotation._();
 
   /// Part after the `@` in annotation.
-  Code get code;
+  @override
+  Code get code => codeForAnnotation;
+
+  /// Use [code] instead.
+  ///
+  /// This property exists in order to keep compatibility with having a concrete
```
natebosch: I don't understand this... This looks like an unused field. Why do we need it?

matanlurey: Ah this is interesting, I can document it better if needed.

Basically `built_value` only generates code for abstract accessors, and because `Expression` implements `code` it "hides" it from `built_value`, and this effectively becomes an empty class.

So I'm redirecting `code` to `codeForAnnotation`. I can make it private though, I think!

matanlurey: Nope, can't make it private (breaks built_value). `@protected` it is.

natebosch: Interesting. Could be worth filing an issue on `built_value` to see if this is a use case we want to support without hacks.

PR #169, opened 2017-11-23
Infer Method.lamba when `null` is passed.

matanlurey:
This prepares the stage for making this the default in 3.0.0.

In reality I don't imagine it will be a breaking change either way, but at least will avoid _requiring_ `true` or `false` in the vast majority of cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,27 @@ void main() {
 }
 ```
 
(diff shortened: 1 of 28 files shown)


matanlurey:
@natebosch: Nice catch, you caught a bug - forgot to check `isStatement`!

Added and fixed!

PR #175, opened 2017-11-26
Make `FunctionType` implement `TypeReference`

### LINKED ISSUE #174: Parameters cannot have a FunctionType type
It doesn't seem possible to write a Method which takes an argument which is a FunctionType because `FunctionType.type` [returns null](https://github.com/dart-lang/code_builder/blob/258fbc8b2aa65eb84a1283b085a4c82c66dc3800/lib/src/specs/type_function.dart#L61)

Failing test:
```dart
  test('should create a method with a funciton type argument', () {
    expect(
        new Method((b) => b
          ..name = 'foo'
          ..requiredParameters.add(new Parameter((b) => b
            ..type = new FunctionType((b) => b
              ..returnType = refer('String')
              ..requiredParameters.add(refer('int')))
            ..name = 'argument'))),
        equalsDart(r'''
          foo(String Function(int) argument)
          '''));
  });
```

natebosch:
Fixes #174

- Add a test which fails before this change and passes after
- Add tests for nested function types since that was untested previously
- Make `FuctionType` implement `TypeReference`, add missing fields
  returning null and rebuild the built_value file.
- Return `this` from the `type` getter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,8 +87,8 @@ void main() {
 * Added `nullSafeProperty` to `Expression` to access properties with `?.`
 * Added `conditional` to `Expression` to use the ternary operator `? : `
 * Methods taking `positionalArguments` accept `Iterable<Expression>`
-* **BUG FIX**: Parameters can take a `FunctionType` as a `type`. `FunctionType`
-  now implements `TypeReference`.
+* **BUG FIX**: Parameters can take a `FunctionType` as a `type`.
+  `Reference.type` now returns a `Reference`.
 
 ## 2.2.0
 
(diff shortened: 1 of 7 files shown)


natebosch:
Hmm, I don't think we can do this. We can't implement both `Built<FunctionType, FunctionTypeBuilder>` and some other built_value class.

I'm not sure what the right way forward is.

matanlurey:
What about just loosening whatever is requesting a `TypeReference` to accept a `Reference`?

natebosch:
Changing `type` to a `Reference` seems to work.

I think, though, that it's a breaking change. Implementations of `SpecVisitor` will be broken by the change in signature of `visitTypeParameters`, and the change of the `type` field is also technically breaking.

matanlurey:
Realistically the only implementation of `SpecVisitor` is `DartEmitter` today.

... so I'm OK with saying the original implementation was broken, and custom implementations of the visitor/emitter will need a few lines updated to keep working.

natebosch:
Sounds good. I added a note to the changelog about the breaking change.

natebosch:
@matanlurey  - if you have a chance to publish soon I'd appreciate it :smile:

matanlurey:
@natebosch Definitely today, I'm back home in about an hour.

PR #187, opened 2018-01-10
Latest build

kevmoo:


pubspec.yaml
```diff
@@ -18,7 +18,8 @@ dependencies:
   meta: ^1.0.5
 
 dev_dependencies:
-  build_runner: '>=0.4.0 <0.7.0'
+  build_config: '^0.2.1'
```
natebosch: why do you need this dep?

kevmoo:
InputSet -> `generateFor: const InputSet(include: const
['lib/src/specs/**.dart']),`

worth exporting?

natebosch:
> worth exporting?

Not sure. I guess this is probably enough of an edge case that it won't matter much. Once we update built_value to expose a build.yaml this file will go away.

matanlurey:
Let's hold off here and wait until we can get `build.yaml` working.

PR #192, opened 2018-01-29
Upgrade to built_value 5.1.0.

davidmorgan:
Update `built_value` and regenerate. This is needed to allow `built_value` to upgrade to the latest `builder_runner`.

The newly generated code provides much better error messages on failure to construct an instance, particularly with nested classes.


matanlurey:
I'll do a release/sync sometime today.

PR #202, opened 2018-02-26
Remove visibleForTesting annotation for 4 functions

srawlins:
These four functions are called from libraries _not_ in `test/` or `testing/` directories:

* `toClosure` - called from lib/src/specs/method.dart
* `InvokeExpression.constOf` - called from lib/src/specs/reference.dart
* `InvokeExpression.newOf` - called from lib/src/specs/reference.dart
* `visitAll` - called from lib/src/specs/expression.dart


matanlurey:
Thanks!

PR #206, opened 2018-04-18
.gitignore and dev_depedency updates

kevmoo:


.gitignore
```diff
@@ -1,14 +1,5 @@
 # Files and directories created by pub
-.buildlog
+.dart_tool
 .packages
-.project
 .pub
```
natebosch: this isn't created anymore is it?

.gitignore
```diff
@@ -1,14 +1,5 @@
 # Files and directories created by pub
-.buildlog
+.dart_tool
 .packages
-.project
```
natebosch: are some of these from IntelliJ or something? Do we want to remove them?

.gitignore
```diff
@@ -1,14 +1,5 @@
...
 .pub
-**/build
-**/packages
-
-# Directory created by dartdoc
-doc/api/
```
natebosch: why remove this?

pubspec.yaml
```diff
@@ -15,15 +15,8 @@ dependencies:
 
 dev_dependencies:
   build: ^0.12.0
-  build_runner: ^0.7.9
+  build_runner: '>=0.7.9 <0.9.0'
```
natebosch: it's a dev dependency and we're already restricted to dart 2 - why do we need to support old build_runner versions?

.gitignore
kevmoo: It is a VERY old thing from Dart Editor days.

.gitignore
kevmoo: packages dir is no longer created
We don't generate docs in this package

pubspec.yaml
kevmoo: Was trying to just support latest. No issues bumping to `^0.8.0`

PR #215, opened 2018-07-19
chore: set max SDK version to <3.0.0

chalin:


pubspec.yaml
```diff
@@ -19,4 +20,4 @@ dev_dependencies:
   built_value_generator: ^5.1.0
```
kevmoo: Bump `build_runner: ^0.9.2`

pubspec.yaml
```diff
@@ -19,4 +20,4 @@ dev_dependencies:
...
   dart_style: ^1.0.0
   source_gen: '>=0.7.5 <0.9.0'
-  test: ^0.12.0
+  test: '>=0.12.0 <2.0.0'
```
kevmoo: Might as well make this `^1.3.0`

pubspec.yaml
chalin: done

pubspec.yaml
chalin: done

PR #224, opened 2018-08-10
Run dartfmt --fix to drop optional new

natebosch:
- Update the SDK constraint to >=2.0.0
- Update README to drop `new` keyword.
- Hold back lib/**/*.g.dart files since they are generated.


natebosch:
ping @matanlurey

PR #227, opened 2018-08-15
Don't emit `new` keyword in generated code

### LINKED ISSUE #220: omit optional new in constructor calls

natebosch:
Closes #220

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 3.1.3
+## 3.1.3-dev.1
 
 * The `new` keyword will not be used in generated code.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.1.3
+
+* The `new` keyword will not be used in generated code.
+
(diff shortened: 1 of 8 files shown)


natebosch:
Yeah releasing as a -dev works for me, want me to change the pubspec?

matanlurey:
If you don't mind, thanks!

natebosch:
Add `-dev.1` in pubpsec and changelog.

Or did you intend for this to be `4.0.0-dev.1`?

matanlurey:
What do you think? Are there any other more visible breaking changes we want to do?

natebosch:
Omitting `const` is tougher but we could try doing that as well.

I'm somewhat more inclined to not call this a breaking change since it's only the more brittle tests that will break - but I don't have strong opinions either way.

matanlurey:
@natebosch Still want this?

natebosch:
Yes, this is still worthwhile.

Renamed version to `3.2.0-alpha` to indicate non-breaking but, give a window for opting in to the alpha and updating goldens.

PR #233, opened 2018-12-07
Use `=` instead of `:` for default values

### LINKED ISSUE #232: Use `=` to separate a named parameter from its default value
This kind of code:

```dart
Method((builder) => builder
  ..returns = const Reference('String')
  ..name = 'getPreamble'
  ..optionalParameters.add(Parameter((param) => param
    ..named = true
    ..type = const Reference('bool')
    ..name = 'minified'
    ..defaultTo = const ToCodeExpression(literalFalse)
  ))
  ..optionalParameters.add(Parameter((param) => param
    ..named = true
    ..type = const Reference('bool')
    ..name = 'shebang'
    ..defaultTo = const ToCodeExpression(literalFalse)
  ))
)
```

Generates:

```dart
String getPreamble({bool minified: false, bool shebang: false})
```

Which triggers a warning from the analyzer:
`Use '=' to separate a named parameter from its default value`

I'm using the `prefer_equal_for_default_values` rule from the Dart linter. It's based on the "Effective Dart" guide: https://www.dartlang.org/guides/language/effective-dart/usage#do-use--to-separate-a-named-parameter-from-its-default-value

code_builder: 3.1.3
Dart VM: 2.1.0
OS: Ubuntu 18.10

natebosch:
Fixes #232

This follows the recommendations in Effective Dart.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.2.0
+
+* Emit `=` instead of `:` for named parameter default values.
+
 ## 3.1.3
 
 * Bump dependency on built_collection to include v4.0.0.
--- a/lib/src/emitter.dart
+++ b/lib/src/emitter.dart
@@ -441,11 +441,7 @@ class DartEmitter extends Object
     }
     output.write(spec.name);
     if (optional && spec.defaultTo != null) {
-      if (spec.named) {
-        output.write(': ');
-      } else {
-        output.write(' = ');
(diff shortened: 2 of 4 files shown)


natebosch:
This does mean the emitted code won't work on old versions of the SDK, but I think we've already decided that we only emit code for Dart 2.0 and above?

matanlurey:
Yup @natebosch!

Issue #221, opened 2018-08-06
omit optional const in const constructor calls


natebosch:
This will be trickier than omitting `new`, because only those which are already in const contexts can be omitted.

matanlurey:
Might need to augment the emitter so it knows where we are currently in a `const` context.

matanlurey:
@natebosch Is it worth this being behind a flag? Is this a breaking change if not?

kevmoo:
@natebosch – see https://github.com/dart-lang/code_builder/pull/238

PR #238, opened 2019-01-29
Omit extra const

kevmoo:
Fixes https://github.com/dart-lang/code_builder/issues/221


kevmoo:
I validated this on both the angular goldens and the build script from build_runner – all good!

PR #242, opened 2019-06-18
Escape newlines in String literals

### LINKED ISSUE #241: Newlines in strings don't work
`literal` and `literalString`, when given a string with a newline, will both leave the newline verbatim *in a single-quoted string*, leading to build errors. This was supposed to be fixed in https://github.com/dart-lang/code_builder/commit/d1d4194a1bbf4a44c315da6e32a62ba085a2fa53 but I'm guessing something's changed since then?

natebosch:
Fixes #241

It is never sensible to write a single quoted string with a newline
character, they should be written as `\n`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.2.1
+
+* Escape newlines in String literals.
+
 ## 3.2.0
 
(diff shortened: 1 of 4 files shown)


refi64:
Thanks!

PR #246, opened 2019-09-13
Remove visibleForTesting from constructor used in lib/src/specs/refer…

srawlins:
…ence.dart

Due to a bug in the analyzer, this is currently not reported. In order to land a fix, I have to clean up violations like this.


natebosch:
What is the error the would be reported?

srawlins:
Sorry, the error is that CodeExpression() is used here:

https://github.com/dart-lang/code_builder/blob/master/lib/src/specs/reference.dart#L116

which is not a test.

srawlins:
Ping

PR #249, opened 2019-11-07
`OR` and `NOT` expressions

### LINKED ISSUE #247: Support for `OR` operator
There is `Expression.and(other)` method which produces an expression which performs logical `AND` operation. However, there is no `Expression.or(other)` method.

I use `code_builder` of version 3.2.0.

ddashenkov:
This pull request adds support for `OR` and `NOT` logical operator expressions.
Previously, there was only an `AND` expression.

For this, we add `Expression.or(Expression)` and `Expression.negate()` methods.

Fixes #247.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: code_builder
-version: 3.2.1
+version: 3.2.2
 
 description: >-
   A fluent, builder-based library for generating valid Dart code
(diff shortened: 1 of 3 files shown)


ddashenkov:
@natebosch, PTAL.

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: code_builder
-version: 3.2.1
+version: 3.2.2
```
kevmoo: Add an entry to the changelog, too!

PR #264, opened 2020-01-12
Add support for Extension methods

nishtahir:


the4thfloor:
@nishtahir your tests are broken

nishtahir:
> @nishtahir your tests are broken

@the4thfloor i'm guessing this is because `extension` isn't a keyword in Dart 2.2 so `dartfmt` doesn't know how to format things correctly. I'm hoping to get some guidance from the team on how to handle this correctly.

the4thfloor:
@nishtahir I guess the best solution would be to use the `@Skip` annotation with a link to an issue to enable these tests when `dartfmt` knows `extensions`

see https://github.com/dart-lang/test/tree/master/pkgs/test#skipping-tests

nishtahir:
It works with the newer toolchain, `2.8.0-dev.0.0`, but fails on `2.2.0 (Tue Feb 26 15:04:32 2019 +0100)`. I'm not sure it's a good idea to skip the tests because it fails on an older version of `dart` that's not going to change.

lib/src/specs/extension.dart
```diff
@@ -0,0 +1,79 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
natebosch: [nit] 2020

lib/src/specs/extension.dart
```diff
@@ -0,0 +1,79 @@
...
+  @nullable
+  Reference get on;
+
+  @override
+  BuiltList<Reference> get types;
+
+  BuiltList<Method> get methods;
+  BuiltList<Field> get fields;
+
+  /// Name of the class.
```
natebosch: [nit] Drop this. It's not a "class", and "Name of the extension." doesn't add any information which isn't clear from context.

lib/src/visitors.dart
```diff
@@ -24,6 +25,8 @@ abstract class SpecVisitor<T> {
 
   T visitClass(Class spec, [T context]);
 
+  T visitExtension(Extension spec, [T context]);
```
natebosch: @matanlurey - is this a breaking change?

the4thfloor: I guess it depends on the Dart version you use. But with the stable Dart 2.7, it should work. If you use an older Dart version, which doesn't know extension methods, the code generator should work but generates unknown extension methods. see the failing tests

natebosch: It's more of a question of whether `implements SpecVisitor` is an encouraged, or even supported, pattern. It doesn't look like this class is public to the package so it might be safe to call this non-breaking.

In general it can be breaking to add a new method to a class if others may use `implements` with that class.

natebosch:
You should be able to bump the minimum SDK version in the `pubspec.yaml` file, and use that newer SDK in `travis.yaml`.

matanlurey:
Closing due to inactivity. Feel free to re-open a PR if/when updated.

PR #273, opened 2020-04-22
Add isNullable field to TypeReference, and isNullSafety to DartEmitter

### LINKED ISSUE #270: Cannot create a TypeReference to a nullable type
If I try to build a TypeReference putting a `?` at the end of `symbol`, it puts it before type arguments.

```dart
TypeReference((TypeReferenceBuilder b) {
    b
      ..symbol = '${type.name}?'
      ..url = _typeImport(type)
      ..types.addAll(type.typeArguments.map(_typeReference));
  });
```

given `type` is the DartType for `List<Foo>`, then this results in `List?<Foo>`.

I don't see a way to create, e.g. `List<Foo>?`.

srawlins:
Fixes #270 

I'm _very_ open on better names to the DartEmitter constructor parameter, `isNullSafety`. The analyzer's LibraryElement has the deprecated name, `isNonNullableByDefault`. WDYT? `hasNullSafety`?

I'd like to have tested a nullable function type, but it didn't seem possible other than setting `symbol` to, e.g. `void Function(int)`.

### DIFF
--- a/lib/src/emitter.dart
+++ b/lib/src/emitter.dart
@@ -59,15 +59,24 @@ class DartEmitter extends Object
   /// lint.
   final bool orderDirectives;
 
+  /// If nullable types should be emitted with the nullable suffix ("?").
+  final bool _isNullSafety;
+
   /// Creates a new instance of [DartEmitter].
   ///
   /// May specify an [Allocator] to use for symbols, otherwise uses a no-op.
-  DartEmitter([this.allocator = Allocator.none, bool orderDirectives = false])
-      : orderDirectives = orderDirectives ?? false;
+  DartEmitter(
+      [this.allocator = Allocator.none,
+      bool orderDirectives = false,
+      bool isNullSafety = false])
+      : orderDirectives = orderDirectives ?? false,
+        _isNullSafety = isNullSafety ?? false;
 
   /// Creates a new instance of [DartEmitter] with simple automatic imports.
-  factory DartEmitter.scoped({bool orderDirectives = false}) {
-    return DartEmitter(Allocator.simplePrefixing(), orderDirectives);
+  factory DartEmitter.scoped(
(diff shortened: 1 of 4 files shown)


srawlins:
CC @natebosch @matanlurey I'd love a review if you have the time.

matanlurey:
Maybe `supportNullSafety: true`?

lib/src/emitter.dart
```diff
@@ -59,15 +59,24 @@ class DartEmitter extends Object
   /// lint.
   final bool orderDirectives;
 
+  /// If nullable types should be emitted with the nullable suffix ("?").
```
natebosch, line 62: Can we expand this a bit? Maybe something like "Null safe syntax should only be enabled if the output will be used with a language version which supports it."

lib/src/emitter.dart
```diff
@@ -59,15 +59,24 @@ class DartEmitter extends Object
...
+  final bool _isNullSafety;
```
natebosch, line 63: How about `useNullSafetySyntax`?

lib/src/specs/type_reference.dart
```diff
@@ -39,6 +39,10 @@ abstract class TypeReference extends Expression
   @override
   BuiltList<Reference> get types;
 
+  /// Optional nullability.
```
natebosch, line 42: Expand this, "An emitter may ignore this if the output is not targeting a language version that supports null safety."

srawlins: Done.

lib/src/emitter.dart
srawlins: Done.

lib/src/emitter.dart
srawlins: Good call, done.

PR #275, opened 2020-04-23
Add 'required' and 'covariant' bools to Parameter

### LINKED ISSUE #274: Parameter needs a bool field for 'required' (also 'covariant')
Mockito's code-generating API will need to be able to declare named parameters as `required`.

srawlins:
Fixes #274

### DIFF
--- a/lib/src/emitter.dart
+++ b/lib/src/emitter.dart
@@ -469,6 +469,13 @@ class DartEmitter extends Object
   }) {
     spec.docs.forEach(output.writeln);
     spec.annotations.forEach((a) => visitAnnotation(a, output));
+    // The `required` keyword must precede the `covariant` keyword.
+    if (spec.required) {
+      output.write('required ');
+    }
+    if (spec.covariant) {
+      output.write('covariant ');
+    }
     if (spec.type != null) {
       spec.type.type.accept(this, output);
       output.write(' ');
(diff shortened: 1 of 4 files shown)


srawlins:
CC @natebosch @matanlurey I'd love a review if you have the time.

lib/src/emitter.dart
```diff
@@ -469,6 +469,13 @@ class DartEmitter extends Object
   }) {
     spec.docs.forEach(output.writeln);
     spec.annotations.forEach((a) => visitAnnotation(a, output));
+    // The `required` keyword must precede the `covariant` keyword.
```
matanlurey, line 472: Do we want to guard this with nullability enabled by the emitter?

I realize it's not 1:1, just wondering.

srawlins: So, I thought of that too, but I saw that there is no prior case of the emitter throwing any exception. And other conditions, like `this.` parameters, says in the comment "Only for constructors", but this is not enforced anywhere. It seems like historically, the policy of code_builder has been "garbage in: garbage out; we're not language spec police."

natebosch: I think it's a difference case from `this.` in constructors.

In the case of `required` it's a language versioning problem. If you are otherwise able to write the exact same code between two target language versions, and only need to omit the `required` like we omit the `?` on nullable types, I think it should be controllable by the emitter.

As it stands without support for skipping this keyword in the ouput, codegen authors will need both conditionals around setting this when building the code, _and_ conditionals for enabling null safe output from the emitter.

srawlins: That makes sense. I just filed https://github.com/dart-lang/code_builder/issues/276

PR #269, opened 2020-03-01
Add missing operators

smkhalsa:
Closes #268
Closes #266

PR #277, opened 2020-05-06
Add "part" directive

### LINKED ISSUE #271: How to build 'part (of)' expression?
Code like `part (of) 'xxx.dart';`, what should i do?

smkhalsa:
fixes #271


test/directive_test.dart
```diff
@@ -39,6 +40,7 @@ void main() {
       equalsDart(r'''
```
smkhalsa, line 40: This test is failing because of issues with whitespace.

smkhalsa: @natebosch any ideas why this may be happening?

smkhalsa: I've skipped this test for now

smkhalsa:
I've rebased on https://github.com/dart-lang/code_builder/pull/269, updated the changelog, and bumped the version.

Can we get this merged, please?

natebosch:
cc @matanlurey - I'm curious for your thoughts on this.

I think we might need to add sorting at some point here. I'm pretty sure a `part` directive can't be followed by an `import` or `export`...

smkhalsa:
closing in favor of #269

smkhalsa:
@natebosch sorting works out of the box with the existing [compare override](https://github.com/smkhalsa/code_builder/blob/part-directive/lib/src/specs/directive.dart#L121) since part is listed last in the `DirectiveType` enum. However, it does require that the `orderDirectives` parameter of `DartEmitter` be set to true.

PR #279, opened 2020-05-18
Sort allocator imports before other directives

natebosch:
Manually specified directive, most likely exports or part directives,
should generally follow imports. Move imports from `refer` URIs to the
beginning so that they get the correct order as long as manually
specified directives where in the correct order, regardless of sorting
directives.

### DIFF
--- a/lib/src/emitter.dart
+++ b/lib/src/emitter.dart
@@ -315,7 +315,7 @@ class DartEmitter extends Object
       }
     }
 
-    final directives = <Directive>[...spec.directives, ...allocator.imports];
+    final directives = <Directive>[...allocator.imports, ...spec.directives];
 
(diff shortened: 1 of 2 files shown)


kevmoo:
We should actually guarantee parts are always after imports/exports – anything else is a syntax error.

natebosch:
> We should actually guarantee parts are always after imports/exports – anything else is a syntax error.

That looks to be the case if you specify that directives should be sorted.

If you don't specify that directives should be sorted the order you'll get is imports from `refer` URIs, then the ones you manually added, in the order you manually added them. As long as you are adding them in the correct order, or using `sortDirectives: true`, it should be fine. This PR solves the problem of having incorrectly ordered directives _even if_ you ordered them correctly manually.

PR #281, opened 2020-05-26
enable and fix a number of lints

kevmoo:


analysis_options.yaml
```diff
@@ -20,15 +20,50 @@ linter:
     - throw_in_finally
 
     # Style Rules
+    - avoid_catching_errors
+    - avoid_private_typedef_functions
+    - avoid_redundant_argument_values
+    - avoid_renaming_method_parameters
+    - avoid_returning_null_for_void
+    - avoid_unused_constructor_parameters
+    - avoid_void_async
```
natebosch, line 29: I guess we don't have any async here anyway, but this is not a lint we should be enabling.

kevmoo: por que?

natebosch: Because performing async work that the caller shouldn't wait for is valid. It should be relatively rare, but I don't see a reason to impose boilerplate around it.

This might get picked up by `pedantic`, which would be fine, but it might also sputter out when we look at enforcement on a large corpus in which case I don't think we should be enforcing it in our packages.

PR #285, opened 2020-06-09
Explicitly use SDK dartfmt

kevmoo:
Trying to fix CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ dart:
 dart_task:
 - test
 - dartanalyzer: --fatal-infos --fatal-warnings .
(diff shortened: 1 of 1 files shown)


natebosch:
Where would it come from otherwise? I wasn't aware of this setting.

kevmoo:
> Where would it come from otherwise? I wasn't aware of this setting.

Just trying to fix https://travis-ci.org/github/dart-lang/code_builder/jobs/696632224#L321-L365

kevmoo:
Oh! It can come from a package that depends on `dart_style` – which code builder does!

PR #287, opened 2020-06-16
Add a thrown getter to Expression

### LINKED ISSUE #286: Cannot throw an expression
I see that `Expression` has getters for `awaited` and `returned`, but not `thrown`. I believe one should be added. The implementation for `returned` invokes a private constructor of BinaryExpression, so I'm not sure how to work-around either...

srawlins:
Fixes #286 

With this change, an expression can be thrown. This can be done as part of an function expression body, or as a statement, by using `.statement`.


srawlins:
Done.

PR #290, opened 2020-07-06
Bump min Dart SDK to 2.7

kevmoo:
Works-around issue with testing Dart 2.6 in CI and changes to
pkg:analyzer

Specifically, it seems there was a change to a non-public analyzer API
that is picked up by packages on Dart 2.6 but the SDK constraint on
`dart_style` prevents picking up a version of that package with the fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: dart
 
 dart:
   - dev
-  - 2.6.0
+  - 2.7.0
 
 dart_task:
 - test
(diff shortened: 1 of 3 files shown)


kevmoo:
Travis log

https://travis-ci.org/github/dart-lang/code_builder/jobs/704428959#L261-L266

```
  Failed to load "test/directive_test.dart":
  Unable to spawn isolate: ../../../.pub-cache/hosted/pub.dartlang.org/dart_style-1.3.3/lib/src/dart_formatter.dart:105:30: Error: Too many positional arguments: 0 allowed, but 1 found.
  Try removing the extra positional arguments.
      scanner.configureFeatures(featureSet);
```

PR #292, opened 2020-07-21
add enum support

### LINKED ISSUE #289: Add support for creating enums
It seems there is no support right now for creating enums. (If i didn't miss anything?)

TarekkMA:
closes #289

### DIFF
--- a/example/example.dart
+++ b/example/example.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:code_builder/code_builder.dart';
+import 'package:code_builder/src/specs/enum.dart';
 import 'package:dart_style/dart_style.dart';
 
 final _dartfmt = DartFormatter();
@@ -10,6 +11,7 @@ final _dartfmt = DartFormatter();
 void main() {
   print('animalClass():\n${'=' * 40}\n${animalClass()}');
   print('scopedLibrary():\n${'=' * 40}\n${scopedLibrary()}');
+  print('jsonEnum():\n${'=' * 40}\n${jsonEnum()}');
 }
 
 /// Outputs:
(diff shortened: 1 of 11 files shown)


TarekkMA:
@matanlurey can you take a look as the PR

example/example.dart
```diff
@@ -3,13 +3,15 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:code_builder/code_builder.dart';
+import 'package:code_builder/src/specs/enum.dart';
```
natebosch, line 6: The examples shouldn't need to import from `src`. Is there something else that needs to be exported?

example/example.dart
```diff
@@ -52,3 +54,32 @@ String scopedLibrary() {
...
+///  @JsonKey('m')
+///  Metric,
+///  @JsonKey('i')
+///  Imperial
+/// }
+/// ```
+String jsonEnum() {
+  final values = <EnumValue>[
+    EnumValue((b) => b
+      ..name = 'Metric'
```
natebosch, line 71: This should be `'metric'` in our examples.

lib/src/specs/enum.dart
```diff
@@ -0,0 +1,97 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
natebosch, line 1: [nit] 2020

TarekkMA: Okay, I have updated the copyright date to 2020

example/example.dart
TarekkMA: What do you mean? Should I rename the enum values to be lowercase?

Is that what should it look like?
```dart
enum Unit {
 @JsonKey('m')
 metric,
 @JsonKey('i')
 imperial
}
```

example/example.dart
TarekkMA: okay this import was useless.

example/example.dart
natebosch: Yes. We always follow best practices from Effective Dart in our examples.

https://dart.dev/guides/language/effective-dart/style#do-name-other-identifiers-using-lowercamelcase

TarekkMA: Oh, that is right, it should be fixed now.

natebosch (approved):

natebosch:
This looks like a great feature. Thanks!

We'll need some internal validation before publishing.

PR #294, opened 2020-08-13
Flip expected and actual in mismatch description

### LINKED ISSUE #293: Wrongly switched actual and expected in `equalsDart` matcher
If you test/run the following snippet:
```dart
import 'package:test/test.dart';

void main() {
  test('test equalsDart',() {
    const actual = Code("final x='x';");
    expect(actual, equalsDart("final x='y';"));
});
}
```

The following error will be generated:
```
Expected: final x='y';
  Actual: StaticCode:<final x='x';>
   Which: final x='x';is different.
          Expected: final x='x';
            Actual: final x='y';
                             ^
           Differ at offset 9
```
Please notice the switch up of `Expected` and `Actual` on the fourth and fifth lines, which should be referring to the same actual and expected as the first and second line. `actual` should be referring to the first parameter of `expect` and not the content of the matcher, since the parameter is also named `actual` and all other matchers confirm to that spec.

This wrong message gets confusing fast, especially when testing larger code generation output where the exact mismatch location is harder to find.

natebosch:
Fixes #293

When describing a mismatch the _expected_ value should be passed to
`equals`, not the actual value. Also remove the source from the start of
the mismatch description since it's already present in the nested
"Actual" field.

Other refactoring:
- Rename `_source` to `_expectedSource` and `result` to `actualSource`
  for clarity.
- Expand the doc comment for `equalsDart`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 3.4.1-dev
+
+* Fix confusing mismatch description from `equalsDart`.
+  https://github.com/dart-lang/code_builder/issues/293
+
 ## 3.4.0
 
 * Introduce `Expression.thrown` for throwing an expression.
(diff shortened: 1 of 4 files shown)


natebosch:
I'll need to test this with some internal test suites before publishing.

PR #297, opened 2020-09-02
Add extension support

smolck:
Continuation of #264.

### DIFF
--- a/lib/code_builder.dart
+++ b/lib/code_builder.dart
@@ -37,6 +37,7 @@ export 'src/specs/expression.dart'
         literalString,
         literalTrue,
         literalFalse;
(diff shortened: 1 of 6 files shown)


matanlurey:
Hi @smolck, thanks for the contribution.

Your code is not following the Dart style guide conventions, so the build has failed. Take a look at some of the other implementations - for example you still use the `new` operator. Try to get your code to match something like `ClassBuilder`, and we can move forward. Thanks!

smolck:
@matanlurey I've updated the code as you requested; is there anything more that needs to be done?

matanlurey:
Sorry for the delay. LGTM, I'm merging!

PR #304, opened 2020-12-21
Migrate to GitHub Actions

athomas:


README.md
```diff
@@ -1,5 +1,5 @@
 [![Pub package](https://img.shields.io/pub/v/code_builder.svg)](https://pub.dev/packages/code_builder)
-[![Build status](https://travis-ci.org/dart-lang/code_builder.svg)](https://travis-ci.org/dart-lang/code_builder)
+[![Build Status](https://github.com/dart-lang/code_builder/workflows/Dart%20CI/badge.svg)](https://github.com/dart-lang/code_builder/actions?query=workflow%3A"Dart+CI")
```
kevmoo, line 2: make sure to refence to the master branch for both the image and the link – or you get weird output

kevmoo (changes requested):

README.md
athomas: Hmm, not sure I understand this. The image points to the default branch (=`master`) for the workflow (the doc says that if you don't specify a branch you get the default branch). The search Link shows all results for the workflow, that seems to be what people are doing in general (often the link is just to the actions tab, without a filter on the workflow or branch, e.g. https://github.com/dart-lang/build/actions).

Or do you mean a completely specific link like this?
https://github.com/dart-lang/build/actions?query=workflow%3A%22Dart+CI%22+branch%3Amaster+event%3Apush

kevmoo: I just pushed the change.

Otherwise you get a green status if the last push to a branch was green, even if it wasn't master!

athomas: > Otherwise you get a green status if the last push to a branch was green, even if it wasn't master!

Not true, but it was on Travis. GitHub Actions badges default to using the default branch (`master` in old repos, `main` in newer/migrated repos) when no specific branch is specified.

See https://github.com/athomas/github-actions-playground, where the badge is green but the latest result is on the always-fail branch.

> [By default, badges display the status of your default branch.](https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/adding-a-workflow-status-badge)

Travis badges don't have a default branch concept, so you needed to specify the master branch to avoid weirdness. Not using the branch explicitly on GitHub Actions makes future default branch migrations (`master` -> `main`) slightly easier.

Issue #217, opened 2018-07-26
Add support for comments - line and doc


kevmoo:

srawlins:
Something to keep in mind for this one: supporting references in doc comments. For example, if I have:

```dart
import 'foo.dart' as _i1;

/// This class extends [Foo].
class Bar extends _i1.Foo {}
```

Then I get a "Only reference in scope identifiers in doc comments" lint report. If a first-class Doc comment could be made, I could add text ("This class extends ") and references (`TypeReference(...)`).

donny-dont:
@srawlins if you take another stab at this my wish list would be

- Add line breaks to support formatting line lengths
- Some markdown support, like adding a link to a URL
- Example code block

srawlins:
Good ideas. I don't foresee mailing a PR any time soon.

devoncarew:
https://github.com/dart-lang/code_builder/pull/389 implements a little bit of this - the ability to have line comments for libraries, emitted at the beginning of the file (`// Generated by ...`).

PR #311, opened 2021-03-08
Support References in docs

srawlins:
Work for #217. (I'm not sure what else is desired for that bug...)

### DIFF
--- a/lib/src/mixins/dartdoc.dart
+++ b/lib/src/mixins/dartdoc.dart
@@ -6,10 +6,10 @@ import 'package:built_collection/built_collection.dart';
 
 abstract class HasDartDocs {
   /// Dart docs.
-  BuiltList<String> get docs;
+  BuiltList<Object> get docs;
 }
 
 abstract class HasDartDocsBuilder {
   /// Dart docs.
-  ListBuilder<String> docs;
+  ListBuilder<Object> docs;
 }
--- a/lib/src/mixins/dartdoc.dart
+++ b/lib/src/mixins/dartdoc.dart
@@ -6,10 +6,16 @@ import 'package:built_collection/built_collection.dart';
 
 abstract class HasDartDocs {
   /// Dart docs.
+  ///
+  /// This list may contain both [String]s and [Reference]s, so that elements
+  /// may be properly referenced with a prefix.
   BuiltList<Object> get docs;
 }
 
 abstract class HasDartDocsBuilder {
   /// Dart docs.
+  ///
+  /// This list may contain both [String]s and [Reference]s, so that elements
+  /// may be properly referenced with a prefix.
   abstract ListBuilder<Object> docs;
 }
--- a/test/specs/class_test.dart
+++ b/test/specs/class_test.dart
@@ -30,7 +30,7 @@ void main() {
     );
   });
 
-  test('should create a class with documentations', () {
+  test('should create a class with documentation', () {
     expect(
       Class(
         (b) => b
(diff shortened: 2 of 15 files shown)


lib/src/mixins/dartdoc.dart
```diff
@@ -6,10 +6,10 @@ import 'package:built_collection/built_collection.dart';
 
 abstract class HasDartDocs {
```
matanlurey, line 7: Let's add comments here and below why `Object` is being used and what a valid type is (and isn't).

test/specs/class_test.dart
```diff
@@ -48,6 +48,26 @@ void main() {
     );
   });
 
+  test('should create a class with documentation with references', () {
+    expect(
+      Class(
+        (b) => b
+          ..name = 'Foo'
+          ..docs.addAll(
```
matanlurey, line 56: Part of me wonders if `docs` should be a more complex class than a `List`, i.e.:

```
class DocsBuilder {
  void addText();
  void addReference();
}
```

... WDUT? It would be more breaking, though.

matanlurey (commented):
Some thoughts. Would love @natebosch to weigh in too!

test/specs/class_test.dart
srawlins: I started out with a much bigger change where I added a new class, `DocComments`, and everything that HasDocComments today also got a `@nullable DocComments` field. Then that class just had a `List<Object>`, so that you could add Strings, and References.

Then I looked back and saw that I had really just created this unnecessary nesting, and confusion with the existing `docs` field. And the emitter would probably have to have some weird logic in it saying "use one or the other." Essentially the new `DocComment` type did not help with the `List<Object>` issue.

If there was a way to hook into BuiltList, with validation logic, that would be tops. Then I could validate the type at runtime. Or add specific methods, `addText` and `addReference`. There isn't, is there? Would I subtype BuiltList/ListBuilder? ☹️

matanlurey: I honestly am not sure. Let's wait and see what @natebosch says - we could also consult with @davidmorgan. 

If you want this change as-is, I'm willing to approve. I do think we should consider making the interface better.

srawlins: Nope! I am definitely in favor of getting this right. And if creating a new field and adding "don't use both" logic to emitter is ultimately the right way, I'm game.

davidmorgan: I think the recommended way to do that would be to make an interface `Doc` and have built value classes `Text` and `Reference` that implement it. I don't think that can be a compatible change, unfortunately.

BuiltList can't do validation, you could validate in the class that has the list, a built_value can do validation in the private `_` constructor, the fields are all initialized when it runs.

lib/src/mixins/dartdoc.dart
srawlins: Done

test/specs/class_test.dart
srawlins: Ack

natebosch:
I do think it would be best to add a new field for this and deprecate the old one.

I think we should move away from `List` and call the new field `doc`, or `docComment` or maybe `documentation` instead of the plural `docs` - and it should be a built value type. We can add a convenience `text` constructor. Usage could migrate from `docs.add('Something.')` to `doc = DocComment.text('Something.')`

srawlins:
Sounds good.

PR #312, opened 2021-03-08
Add genericClosure on Method

### LINKED ISSUE #309: Add an API to convert a Method to a function literal with its type parameters
If I define a Method like:

```dart
Method((b) {
    var typeT = TypeReference((b) => b..symbol = 'T');
    b.types.add(typeT);
    b.requiredParameters.add(Parameter((b) => b..name = 'arg'..type = typeT));
    b.body = Code('');
});
```

then `Method.closure` returns `(T arg) {}`. This is invalid Dart code as there is no `T` type in scope. I want it to return the generic function literal: `<T>(T arg) {}`.

I'm sure there is a reason for clearing the types in `toClosure`, so I suspect a new API is needed. :( `Method.closure` is a getter, not a function, so we can't add an optional parameter there... Something like `toGenericClosure`? Happy to contribute.

srawlins:
Fixes #309 

The existing `closure` getter strips a Method of its return type and type parameters. This new API, `genericClosure` keeps the type parameters, as they are legal, and can be useful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 3.7.0-dev
+
+* Add support for converting a Method to a generic closure, with
+  `Method.genericClosure`.
+
 ## 3.6.0
 
 * Add support for creating `extension` methods.
--- a/lib/src/specs/expression/closure.dart
+++ b/lib/src/specs/expression/closure.dart
@@ -4,12 +4,21 @@
 
 part of code_builder.src.specs.expression;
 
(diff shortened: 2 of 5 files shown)


matanlurey (approved):

srawlins:
I'll add a changelog entry and bump version.

srawlins:
@matanlurey does `3.7.0-dev` sound right?

matanlurey (approved):

PR #317, opened 2021-04-02
Migrate to null safety

jakemac53:
One weird thing here, the `FunctionType` class implements `Reference` which has a non-nullable `symbol` field, but it was overriding it with a getter that returns `null`. I changed that to instead throw an `UnsupportedError` which is a bit funky but makes it statically safe and does make the tests happy. Open to suggestions on that.

### DIFF
--- a/lib/src/allocator.dart
+++ b/lib/src/allocator.dart
@@ -51,7 +51,7 @@ class _Allocator implements Allocator {
   @override
   String allocate(Reference reference) {
     if (reference.url != null) {
-      _imports.add(reference.url);
+      _imports.add(reference.url!);
     }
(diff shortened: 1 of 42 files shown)


lib/src/allocator.dart
```diff
@@ -51,7 +51,7 @@ class _Allocator implements Allocator {
   @override
   String allocate(Reference reference) {
     if (reference.url != null) {
-      _imports.add(reference.url);
+      _imports.add(reference.url!);
```
natebosch, line 54: [nit] I would pull out a `url = reference.url` variable for this and below

natebosch (approved):

PR #318, opened 2021-04-06
Null safety

mpa1234567:
Sorry to be late! This was my null safety migration.
Maybe you can still use some of the code like the new features for fields (late and abstract)
All the code looks just like the same. But how could you generate the built values without put //@dart=2.12 in all sources?
Well. Hope I put it faster next time but (I was testing it)...


natebosch:
Thanks for the PR!

I have merged support for `late` fields. I'll take a look at `abstract` fields too.

Issue #299, opened 2020-10-01
Default DartEmitter constructor has confusing optional parameters


matanlurey:
We might want to deprecate it, add one using named parameters, and remove it:

```dart
  DartEmitter(
      [this.allocator = Allocator.none,
      bool orderDirectives = false,
      bool useNullSafetySyntax = false])
      : orderDirectives = orderDirectives ?? false,
        _useNullSafetySyntax = useNullSafetySyntax ?? false;
```

PR #319, opened 2021-04-06
change the DartEmitter constructor to use named parameters

jakemac53:
Fixes https://github.com/dart-lang/code_builder/issues/299

Going with a one-shot approach to keep the same constructor name, it looks feasible to migrate all at once internally and we have a breaking change for NNBD anyways.

### DIFF
--- a/lib/src/emitter.dart
+++ b/lib/src/emitter.dart
@@ -71,9 +71,9 @@ class DartEmitter extends Object
   ///
   /// May specify an [Allocator] to use for symbols, otherwise uses a no-op.
   DartEmitter(
-      [this.allocator = Allocator.none,
+      {this.allocator = Allocator.none,
       bool? orderDirectives = false,
-      bool? useNullSafetySyntax = false])
+      bool? useNullSafetySyntax = false})
       : orderDirectives = orderDirectives ?? false,
         _useNullSafetySyntax = useNullSafetySyntax ?? false;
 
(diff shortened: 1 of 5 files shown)


lib/src/emitter.dart
```diff
@@ -71,17 +71,19 @@ class DartEmitter extends Object
   ///
   /// May specify an [Allocator] to use for symbols, otherwise uses a no-op.
   DartEmitter(
-      [this.allocator = Allocator.none,
+      {this.allocator = Allocator.none,
       bool? orderDirectives = false,
-      bool? useNullSafetySyntax = false])
+      bool? useNullSafetySyntax = false})
```
natebosch, line 76: I can't remember if we had already discussed this - why do these `bool` need to be nullable if we have a default?

kevmoo: Agreed! run w/ the breaking change!

jakemac53: For unsound apps, but I am OK with making this change.

natebosch: FWIW I'd rather do `useNullSafetySyntax = `(useNullSafetySyntax as dynamic) ?? false;` than to put the nullable type in the signature.

jakemac53: @natebosch then I have to cast it back to bool also... what about just not supporting unsound apps that pass an explicit null? Imo for this package that would be fine.

natebosch: yeah, usage is low and we can reasonably fix opted out usage before publishing with an expanded constraint. I'm in favor of the behavior change.

natebosch (approved):

PR #322, opened 2021-04-19
Add a parenthesized expression

### LINKED ISSUE #305: asA generates code which prevents an expression from being an arrow function
I added this test case to `test/specs/code/expression_test.dart` around line 470:

```dart
test('should emit an explicit cast in an arrow function', () {
    expect(
      refer('callback').call([
        Method((b) => b..body = refer('foo').asA(refer('String')).code).closure
      ]),
      equalsDart('callback(() => ( foo as String ))'),
    );
  });
```

The test fails because the output is actually `callback(() { ( foo as String )...` which

a. does not have the expected `=>`, and
b. does not include a semi-colon, so the output is not even parseable Dart.

natebosch:
Fixes #305

Add a `ParenthesizedExpression` with special handling for the output
instead of using a `Block` to avoid influencing the behavior of other
calls that vary based on whether an argument is an expression or a
block.

### DIFF
--- a/test/specs/method_test.dart
+++ b/test/specs/method_test.dart
@@ -67,6 +67,18 @@ void main() {
     );
   });
 
+  test('should create a lambda method if the value is cast', () {
+    expect(
+      Method((b) => b
+        ..name = 'returnsCastedValue'
+        ..returns = refer('Foo')
+        ..body = refer('bar').asA(refer('Foo')).code),
+      equalsDart(r'''
+        Foo returnsCastedValue() => (bar as Foo)
+      '''),
+    );
+  });
+
   test('should create a normal method implicitly', () {
     expect(
       Method.returnsVoid((b) => b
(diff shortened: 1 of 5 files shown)


srawlins (approved):
Thanks!

test/specs/method_test.dart
```diff
@@ -67,6 +67,18 @@ void main() {
...
   });
 
+  test('should create a lambda method if the value is cast', () {
+    expect(
+      Method((b) => b
+        ..name = 'returnsCastedValue'
+        ..returns = refer('Foo')
+        ..body = refer('bar').asA(refer('Foo')).code),
+      equalsDart(r'''
+        Foo returnsCastedValue() => (bar as Foo)
```
jakemac53, line 77: should we be expecting a semicolon here or does that get added by something else?

jakemac53 (approved):

test/specs/method_test.dart
natebosch: The semicolon gets added by something else.

PR #323, opened 2021-04-20
CI: use dart setup

kevmoo:


kevmoo:
Feel free to squash/merge and delete the branch if you're happy

.github/workflows/test-package.yml
```diff
@@ -20,17 +20,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sdk: [dev]
-        version: [latest]
-        include:
-          - sdk: stable
-            version: 2.12.0
+        sdk: [2.12.0, dev]
```
natebosch, line 23:
```suggestion
        sdk: [dev]
```

We don't want to have the formatting check on both SDK.

I had been configuring to do analyzer on both, but I think I've convinced myself since that it is not worth it..

PR #332, opened 2021-06-21
Introduce `Expression.spread`

donny-dont:
Add support for the spread operator `...` when using collections.


natebosch (approved):

donny-dont:
Thanks @natebosch is a variant for `?...` of use as well?

natebosch:
> Thanks @natebosch is a variant for `?...` of use as well?

Yeah I think a `nullAwareSpread` could be worthwhile.

PR #335, opened 2021-06-23
Add annotations to libraries

donny-dont:
Adds the ability to annotate a `Library` generated by `code_builder`. This is to support scenarios like JavaScript interop using `package:js` where the library itself needs to be annotated with `@JS()`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Add support for `Expression.nullChecked` to add a null assertion operator.
 * Add support for creating `mixin`s.
 * Add `Expression.nullSafeSpread` for the null aware spread operator `...?`.
+* A `Library` can now be annotated.
 
 ## 4.0.0
 
--- a/lib/src/emitter.dart
+++ b/lib/src/emitter.dart
@@ -389,6 +389,13 @@ class DartEmitter extends Object
       }
     }
 
+    if (spec.name != null) {
+      spec.annotations.forEach((a) => visitAnnotation(a, output));
+      output.write('library ${spec.name!};');
+    } else if (spec.annotations.isNotEmpty) {
+      throw StateError('a library name is required for annotations');
+    }
+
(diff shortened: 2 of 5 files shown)


natebosch:
I'm a little hesitant on this one.

Some of our tools allow annotating the first import and treats it as if it was an annotation on the library, but that isn't really the way the grammar works as I understand it. If any tools _aren't_ taking the extra step of reading annotations on the first import this feature will appear misleading.

The slightly more correct way to annotate a library is to have a `library` statement and a library name. We don't support either in `code_builder`.

@jakemac53 - WDYT?

If we had https://github.com/dart-lang/language/issues/1073 this would be easy - we could always add a `library;`.

We could potentially add this feature for now with a comment calling out that support may be inconsistent, and then update it if we get that language feature.

donny-dont:
Would you be against `Library` optionally having a name then?

In this case if you add an annotation but not a `name` then it could `throw` and then later if the language allows a nameless `library` declaration and you're targeting a version of Dart with that it removes the restriction.

jakemac53:
Ya I think we should require a named library from the user if we want to allow annotating it.

donny-dont:
@jakemac53 @natebosch should I also mixin `Library` with `HasDartDocs` and if a `name` isn't present `throw` as well while I'm here?

donny-dont:
Ok I just added the `annotations` part so I think this is good to review.

donny-dont:
@natebosch @jakemac53 the library name is now generated. Happy to add documentation as well if that's something that you all want. Wasn't sure since #311 implied documentation might get a revamp.

jakemac53:
This generally LGTM but I will let Nate have the final say as he is more involved in this package

donny-dont:
@natebosch any thoughts on this?

natebosch (approved):

natebosch:
@donny-dont - sorry for the delay. This is published now.

PR #336, opened 2021-06-24
Introduce `Expression.nullSafeSpread`

donny-dont:
Add the null aware variant of the spread operator.

### DIFF
--- a/lib/src/specs/expression.dart
+++ b/lib/src/specs/expression.dart
@@ -303,6 +303,14 @@ abstract class Expression implements Spec {
         addSpace: false,
       );
 
+  /// This expression precenede by the null safe spread operator `?...`.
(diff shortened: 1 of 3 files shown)


donny-dont:
Changelog rebased @natebosch

lib/src/specs/expression.dart
```diff
@@ -303,6 +303,14 @@ abstract class Expression implements Spec {
         addSpace: false,
       );
 
+  /// This expression precenede by the null safe spread operator `?...`.
+  Expression get nullSafeSpread => BinaryExpression._(
+        const LiteralExpression._('?...'),
```
natebosch, line 308: I think we want `...?` not `?...` right?

donny-dont: Fixed

natebosch (approved):

PR #346, opened 2021-11-07
Add bitwise operators

m0nac0:
Adds bitwise and, bitwise or, bitwise xor operators inluding tests. These were missing so far.


m0nac0:
I hereby declare the contents of the two commits in this PR public domain. If anyone else wants to sign the CLA and contribute the bitwise operators, that'd be perfectly fine with me.

PR #352, opened 2022-03-21
Ignore lint about private import prefixes

natebosch:
In generated code is is more important to avoid potential name
conflicts, even unlikely ones, than to satisfy style lints since the
code is rarely seen. Ignore
`no_leading_underscores_for_library_prefixes` since this lint is in the
recommended set in `package:lints`.


natebosch:
I expect some CI to start failing after publishing this since we have hardcoded expectations of `code_builder` output in a few places. I'm not planning on a major version bump - when we wrote those tests we did so with the expectation that they might be a little fragile.

I will run this through an internal presubmit before landing to make sure there won't be difficult getting it rolled.

jakemac53:
> conflicting with string interpolation in a few places.

Is this breaking potentially?

jakemac53 (approved):

jakemac53:
> My current thinking is to land this and try to land fixes for the tests atomically with the roll. WDYT

sgtm

rakudrama:
My thought here is that if you are generating code, make it as robust as possible against conflicts in imports, which means using private identifiers as much as possible.
I don't fully understand the motivation for the lint, but it is driving the code in the direction of being more fragile, so I would just suppress it in generated code.

jakemac53:
> I don't fully understand the motivation for the lint,

TBH I don't understand the motivation either

natebosch:
I'd be OK with just writing an `// ignore_for_file`.

I'm a little bit worried about other folks piling on with more lints they'd like ignored in this output by default, but I think this one is OK to do since it's in the recommended set of lints.

jakemac53 (approved):

PR #353, opened 2022-03-25
Add a namedRequiredParameters field to FunctionType

### LINKED ISSUE #337: No way to create a FunctionType with a required named parameter
I can create a FunctionType with a named parameter like so:

```dart
FunctionType((b) {
  b
      ..returnType = refer('int')
      ..requiredParameters.add(refer('String'))
      ..namedParameters['p1'] = refer('double');
}); // => int Function(String, {double p1});
```

But there is no way to make that named parameter required; no way to generate `int Function(String, {required double p1})`. I imagine this would require a breaking change. 😦 

This is required for mockito. If a user has a class:

```dart
class C {
  void f(void Function({required int i}) p) {}
}
```

then it is impossible to generate a subclass of C with a legal overriding `f`.

srawlins:
Fixes #337


srawlins:
CC @natebosch. #337 is blocking some at least one migration to null safety, as mockito cannot generate a proper override without being able to specify required named parameters in a function type.

natebosch:
I'm worried that if we get asked to implement support for default values or something similar we're going to get stuck.

Should we consider designing a fully general `Parameter` class which is fully general instead of adding another version of the behavior we have?

natebosch:
cc @jakemac53

srawlins:
I think a fully general `Parameter` is a good idea. I'm less comfortable designing that API, as it would (I think) be breaking, and tricky/opinionated to introduce it in a incremental-migration way.

jakemac53:
> I'm less comfortable designing that API, as it would (I think) be breaking, and tricky/opinionated to introduce it in a incremental-migration way.

In general we can do a non-incremental migration with this package. Even internally I think a one-shot migration with the roll should be feasible.

natebosch:
Hmm, was I mistaken about this? Maybe it's not sensible to worry about default values in function types - that only matters in actual method definitions where we already have more flexibility.

We do already have a `Parameter` class for `Method`. I think the design there is still not optimal - it separates `optionalParameters` from `requiredParameters`, even though that optional or not information is also included in the `Parameter` object.

I'm tempted to try deprecating all parameter related fields on both `Method` and `FunctionType` and replacing them with a single `List<Parameter>` field. When we emit code we could either combine them, or emit an error if you use both.

jakemac53:
Ya I would say the existing API for methods also needs a revamp. I would separate `named` and `positional` parameters personally into separate fields (but share the same `Parameter` class). Then I would just add a generic list of modifiers to `Parameter` which is more forward compatible with the language for any future modifiers.

srawlins:
> I'm tempted to try deprecating all parameter related fields on both Method and FunctionType and replacing them with a single List<Parameter> field. When we emit code we could either combine them, or emit an error if you use both.

+1. This gives confidence that named and required-named remain in the order specified, in case the developer prefers an order there (not that it matters a lot when specifying parameters; just for presentation-to-the-dev purposes).

natebosch:
@srawlins - did you have trouble regenerating the built_value files? It looks like the `tool/builder.dart` is written against an older version of `built_value`, and there are some other errors.

srawlins:
Oh yeah I should have mentioned that. I tried for like 30 minutes to regenerate, even following the steps (which I think I wrote) in the README to regenerate, and it always failed on static analysis errors. :( So I did it by hand.

natebosch:
@srawlins - we might need a bit more time to think about a way to unify `FunctionType` and `Method` parameter lists. I have some thoughts I'll try to capture in an issue soon. 

Should we go forward with this PR to get you unblocked in the mean time? It would mean likely we'd end up deprecating this new field with the others if/when we have a unified solution, but that's probably not a big deal.

natebosch (approved):

PR #358, opened 2022-05-21
Add enhanced enums support

dnys1:
Adds support for the enhanced enums language feature by extending the properties of the current `Enum` class


natebosch:
Thanks for the PR, sorry for the delay in review. I tweaked the naming a bit.

I'll run some internal validation, then I think we should be able to merge this.

@srawlins - can you also take a pass at a PR review on this?

dnys1:
Sounds great and no worries. Thanks, Nate!

test/specs/enum_test.dart
```diff
@@ -73,4 +73,312 @@ void main() {
...
+
+  test('should create an enum which targets a redirecting factory constructor',
+      () {
+    final myEnum = Enum((b) => b
+      ..name = 'MyEnum'
+      ..constructors.addAll([
+        Constructor((c) => c..constant = true),
+        Constructor((c) => c
+          ..constant = true
+          ..factory = true
```
srawlins, line 171: Oddly, const factory constructors are not allowed. See in [the spec](https://github.com/dart-lang/language/blob/master/accepted/future-releases/enhanced-enums/feature-specification.md):

> No-one is allowed to invoke a generative constructor and create an instance of the enum other than the enumerated enum values. That also means that a redirecting factory constructor cannot redirect to a generative constructor of an enum, and therefore no factory constructor of an enum declaration can be const, because a const factory constructor must redirect to a generative constructor.

I think we can keep the body of this constructor, unless you want to change it to something legal (could just always return `a` or throw or something...

srawlins (approved):
I think this looks great. The value-as-constructor-call syntax is so weird, but the tests show that it all works. Generics, arguments, named constructors. I don't think this needs anything else.

test/specs/enum_test.dart
dnys1: Interesting - good catch! @natebosch - what are your thoughts?

natebosch: I think it's probably fine to leave the test as is. I'm not overly worried about only demonstrating correct code in the tests, and there are plenty of ways we allow outputting invalid code.

PR #363, opened 2022-06-07
Add declareVar/Final/Const

### LINKED ISSUE #343: `Expression.assign` is confusing: Either documentation or output is wrong
Given the following code

```dart
import "package:code_builder/code_builder.dart";

void main() {
  final emitter = DartEmitter();

  final e1 = refer("Hello").newInstance([]).assignFinal("hello");
  final e2 = refer("Hello").newInstance([]).assign(refer("hello"));

  print(e1.accept(emitter));
  print(e2.accept(emitter));
}
```

Personally, I would expect it to print:

```dart
final hello = Hello()
hello = Hello()
```

As I'm essentially `assign`ing an expression to a variable reference.

And, in fact, [the documentation for `Expression.assign`](https://pub.dev/documentation/code_builder/latest/code_builder/Expression/assign.html) states:

> Return `{other} = {this}`.

where `{other}` is `hello` and `{this}` is `Hello()`.

However, the above code produces the output like so:

```dart
final hello = Hello()
Hello() = hello
```

Is this a bug in the package or a documentation error?

Like I said, personally, it makes sense for it to output the former. If the actual output here (the latter) is the intended output, please update the documentation (I can of course create a PR). And also, would adding `Expression.assignTo`, which produces the former, be a possibility (creating a new issue of course)?

Thanks!

natebosch:
Closes #343

These utilities serve as a replacement for `assignVar/Final/Const` which
have an inverted responsibility - the variable is told about it's
assignment, rather than an assignment being told about the variable it
is assigned to. Unifies the interaction with `Expression.assign`.

Use named instead of positional optional arguments. Add an optional
argument for a `late` prefix for the final and var declarations. Const
variables may not be `late`.

Will deprecate the `assign*` methods in a followup after doing some
cleanup of internal usage.


lib/src/specs/expression.dart
```diff
@@ -323,6 +323,46 @@ abstract class Expression implements Spec {
...
+///
+/// Returns `const {variableName}`, or `const {type} {variableName}`.
+Expression declareConst(String variableName, {Reference? type}) => type == null
+    ? LiteralExpression._('const $variableName')
+    : BinaryExpression._(
+        const LiteralExpression._('const'), _typedVar(variableName, type), '');
+
+/// Declare a variable named [variableName].
+///
+/// Returns `var {variableName}`, or `{type} {variableName}`.
```
jakemac53, line 336:
```suggestion
/// Returns `final {variableName}`, or `final {type} {variableName}`.
```

jakemac53 (approved):

PR #368, opened 2022-07-22
Add support for super initializer constructor arguments

yanok:


srawlins:
@jakemac53 would you be able to review this, or are you comfortable with me reviewing it?

lib/src/specs/method.dart
```diff
@@ -180,6 +180,12 @@ abstract class Parameter extends Object
   /// This is only valid on constructors;
   bool get toThis;
 
+  /// Whether this parameter should be passed to super\
+  /// constructor (i.e. `super.`).
+  ///
+  /// This is only valid on constructors;
+  bool get toSuper;
```
jakemac53, line 187: A bit of a weird name, but I suppose it is consistent with `toThis`?

jakemac53 (commented):
This generally LGTM but we also need an update to the version and an entry in the changelog.

jakemac53:
If you don't mind also removing the `invariant_booleans` lint from the `analysis_options.yaml` file, that will get the rest of the tests running.

lib/src/specs/method.dart
yanok: Yep, I've just copied the naming scheme from `toThis`.

yanok:
> This generally LGTM but we also need an update to the version and an entry in the changelog.

I added an entry to the changelog but do you really bump the version for every change? Or did I understand it incorrectly?

yanok:
> If you don't mind also removing the `invariant_booleans` lint from the `analysis_options.yaml` file, that will get the rest of the tests running.

I disabled the lint, but now it fails with 2.12 version...

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: code_builder
-version: 4.2.0-dev
+version: 4.2.1-dev
```
jakemac53, line 2: Looks like we were already on a dev version so you can keep the number the same

```suggestion
version: 4.2.0-dev
```

jakemac53:
> I added an entry to the changelog but do you really bump the version for every change? Or did I understand it incorrectly?

What you added is fine - we do want to list all user visible changes in the changelog. But see my note on the number, since were are already in a dev release you can keep that same dev release version.

> I disabled the lint, but now it fails with 2.12 version...

I am not fully understanding the output of that failing tests, but super parameters were not supported until 2.17.0. We may want to bump the min SDK for this package to that version? That probably has something to do with the failure.

yanok:
> What you added is fine - we do want to list all user visible changes in the changelog. But see my note on the number, since were are already in a dev release you can keep that same dev release version.

Yes, that's what I thought about the dev version. Rolled back the version bump.

> I am not fully understanding the output of that failing test, but super parameters were not supported until 2.17.0. We may want to bump the min SDK for this package to that version? That probably has something to do with the failure.

It was surprising that it fails, since the test just generates the text, but then I realized we probably run `dartfmt` on the generated code, it breaks since it can't parse super parameter, so the code stays unformatted and that breaks the test. I bumped the minimal SDK version but it's a pity we have to do this just for tests to work.

jakemac53 (approved):

jakemac53:
cc @srawlins should we go ahead and release this package or are we waiting on anything else?

srawlins:
Not waiting on anything else, I think

PR #373, opened 2022-08-05
Deprecate assignVar, assignFinal, assignConst

natebosch:
Point to the relevant `declare*` API which replaces them.


kevmoo (approved):
🤷 :shipit:

Issue #365, opened 2022-07-05
Option to generate trailing commas after lists of parameters


hacker1024:
As far as I'm aware, there's no way to generate trailing commas at the end of constructor/function invocations/declarations. It looks like the emitter is hard-coded to leave them out.

It'd be great if a feature to include them was added. I'm generating code for humans to read, so the styling is important to me.

This could be enabled on a global scale (like the `orderDirectives` option), or on a local scale as an option in the constructor/function/invocation classes.

hacker1024:
A very quick-and-dirty proof-of-concept, adding the exact `writeTrailingNamedParameterCommas` option that I needed can be found here: https://github.com/hacker1024/code_builder/commit/7fa1ac26e526763e4cc70c4345b514bb830c78c8

Pierre2tm:
Very interested to have this feature !

srawlins:
@jakemac53 @natebosch would this be accepted perhaps as a flag on Emitter? Seems like a reasonable place to put it as an opt-in.

This is important now for performance reasons. Adding trailing commas can improve `dart format` performance by 10x in some cases.

(I guess I don't _have_ to format my generated code but gosh darn it I love formatted code and I don't want to open the floodgates of writing my own HalfDecentFormatter in mockito just to avoid 1000 column lines and unreadable 0-indented swaths of code just so that debugging isn't a nightmare.)

srawlins:
To get into some details:

* I think we should not distinguish between argument lists, parameter lists, collection literals, (any other legal locations). Just one option to add trailing commas.
* The [require_trailing_commas](https://dart-lang.github.io/linter/lints/require_trailing_commas.html) lint rule suggests using trailing commas when a list doesn't fit in one line, but we don't really have that luxury. I suggest instead we use a simple rule that we either use a trailing comma for any list with more than 1 element, or for any list at all. In the latter case, I think code could get unreadable quickly. I'd much prefer to add a trailing comma for any list with multiple elements.

natebosch:
Is the performance differential with trailing commas new?

cc @munificent 

@srawlins - Is there any reason not to make this the default and only option?

munificent:
> Is the performance differential with trailing commas new?

Argument lists have always been relatively slow to format because:

1. The number of ways an argument list can split is linear in the number of arguments.
2. The way an argument list splits can affect both the way the argument expressions split, and the way surrounding code splits, since the `)` is attached to the last argument.

[...]

PR #376, opened 2022-08-17
Add support for trailing commas in Emitter

srawlins:
Fixes https://github.com/dart-lang/code_builder/issues/365

This adds trailing commas in various places according to the following rules:

* in a collection literal if there are more than one element
* in an argument list if there are more than one argument (positional or named)
* in a constructor parameter list, method parameter list, or function type parameter list, if there are more than one parameter (positional, optional, or named)

The last one was certainly the most complex, because of the optional/named delimiters (`[]`/`{}`).
* A parameter list with one positional or one optional or one named parameter does _not_ get a trailing comma.
* A parameter list with multiple positional parameters and no optional or named parameters gets a trailing comma.
* A parameter list with at least one optional parameter or at least one named parameter, and at least two parameters in total, gets a trailing comma before the closing delimiter (`]` or `}`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 * Only emit `late` keyword when using null safety syntax.
 * Use implicit `const` when assigning to a `declareConst` variable.
 * Deprecate `assignVar`, `assignConst`, and `assignFinal`.
+* Add trailing commas to any parameter list, argument list, or collection
+  literal which has more than one element.
 
 ## 4.2.0
 
--- a/lib/src/emitter.dart
+++ b/lib/src/emitter.dart
@@ -68,33 +68,22 @@ class DartEmitter extends Object
   /// a Dart language version which supports it.
   final bool _useNullSafetySyntax;
 
-  /// If trailing commas should be used where they are allowed.
-  ///
-  /// If `true`, an argument list, parameter list, or collection literal with
-  /// more than one element will end with a trailing comma.
-  final bool _useTrailingCommas;
-
   /// Creates a new instance of [DartEmitter].
   ///
   /// May specify an [Allocator] to use for symbols, otherwise uses a no-op.
   DartEmitter(
       {this.allocator = Allocator.none,
       this.orderDirectives = false,
-      bool useNullSafetySyntax = false,
-      bool useTrailingCommas = false})
-      : _useNullSafetySyntax = useNullSafetySyntax,
-        _useTrailingCommas = useTrailingCommas;
+      bool useNullSafetySyntax = false})
+      : _useNullSafetySyntax = useNullSafetySyntax;
 
   /// Creates a new instance of [DartEmitter] with simple automatic imports.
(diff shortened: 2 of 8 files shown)


natebosch:
I'm nervous about sending the signal that `code_builder` is intended to make "pretty" output, although nothing immediately comes to mind where we might need to have ugly output.

I would prefer if we err on the side of being opinionated over being configurable. If we think trailing commas is a better default I'd like to use it everywhere, rather than increase the API surface. WDYT @jakemac53 ?

natebosch:
I guess one other area that may come up is that we explicitly _don't_ want to add options to have output that conforms to style lints. I would not accept PRs that, for instance, have options to conform to prefer single or double quotes, or to sort members within a class in a certain order.

srawlins:
> I would prefer if we err on the side of being opinionated over being configurable.

I don't mind this idea. I thought there might be a question of breakage (internal goldens?) but if code_builder has some contract that we can change the textual content of the output code, then that's great.

> I guess one other area that may come up is that we explicitly don't want to add options to have output that conforms to style lints.

Yeah my motivation here is 100% for the performance concerns in dart_style. I took the effort to make a distinction between 1-element lists and multiple-element lists to make the generated output _bearable_, for developers writing builders, and users ctrl-clicking into generated code. But if it comes up against a principle, I'm find just adding commas anywhere.

sigurdm:
Maybe it would be worth considering some recursive logic. If the single argument of an invocation itself uses trailing comma, then it should probably do as well...

Compare:
```dart
    foo(bar(
      takes,
      a,
      lot,
      of,
      arguments,
    ));
```
to
```
    foo(
      bar(
        takes,
        a,
        lot,
        of,
        arguments,
      ),
    );
```
I much prefer the second...

jakemac53:
I am fine with just always using them, at least for argument lists and list/map/set literals?

The place I really dislike the style is for parameter lists.

srawlins:
I've updated the code to always consider adding trailing commas. Trailing commas are still only added if there are multiple elements in a list. I deleted the tests I added as I think everything is covered with existing tests.

This feature still applies to parameter lists; let me know if it should not.

jakemac53:
> This feature still applies to parameter lists; let me know if it should not.

I think it's fine, I will survive lol. Technically parameter lists could also get quite large, and bob also mentioned that if you have arrow body functions with a large expression, trailing commas in parameter lists will help.

natebosch (approved):
Can you add a CHANGELOG?

natebosch:
> I thought there might be a question of breakage (internal goldens?) but if code_builder has some contract that we can change the textual content of the output code, then that's great.

Yeah we have made changes to the specific output in the past - we consider a test on the specific content output to be opting in to needing updates.

It will make the next package roll more annoying since we will have to atomically update any golden tests, but that's typically pretty manageable.

jakemac53 (approved):

srawlins:
Changelog added; thanks!

Issue #381, opened 2022-12-03
Unclear how to generate an anonymous function


eseidel:
It's not clear (from examples or unit tests) how one might generate an anonymous function for passing to a call, e.g.:

```
setState(() {
  doSomething();
})
```

Sure I could use a Code() block, but I'm looking for the Method() variant.  e.g.

```
refer('setState').call([
// What do I put here?
]);
```

FunctionType is allowed there (its an expression), where as Method is not, but I suspect I actually want a Method, maybe just one without a name?  And then wrap it in a CodeExpression?

eseidel:
Found it.  Method().closure it the magic.

https://github.com/dart-lang/code_builder/blob/master/test/specs/method_test.dart presumably needs a test, maybe I'll add one.

PR #384, opened 2022-12-06
Add an example of generating an anonymous closure

eseidel:
This fixes https://github.com/dart-lang/code_builder/issues/381.

### DIFF
--- a/test/specs/method_test.dart
+++ b/test/specs/method_test.dart
@@ -590,4 +590,19 @@ void main() {
       '''),
     );
   });
+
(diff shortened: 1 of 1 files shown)


eseidel:
Amusingly if you leave off the ..body for Method it just produces '(a)' which is probably a separate bug.   Also Code('') produces '{ }', note the space, which may also be a bug.  🤷

natebosch (approved):

natebosch:
Thanks!

PR #387, opened 2022-12-14
add support for emitting typedefs

devoncarew:
- add support for emitting typedefs

### DIFF
--- a/lib/src/specs/typedef.dart
+++ b/lib/src/specs/typedef.dart
@@ -1,0 +1,55 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:built_collection/built_collection.dart';
+import 'package:built_value/built_value.dart';
+import 'package:meta/meta.dart';
+
+import '../base.dart';
+import '../mixins/annotations.dart';
+import '../mixins/dartdoc.dart';
+import '../visitors.dart';
+import 'expression.dart';
+
+part 'typedef.g.dart';
+
+@immutable
+abstract class TypeDef extends Object
+    with HasAnnotations, HasDartDocs
+    implements Built<TypeDef, TypeDefBuilder>, Spec {
+  factory TypeDef([void Function(TypeDefBuilder)? updates]) = _$TypeDef;
+
+  TypeDef._();
+
+  /// Name of the typedef.
+  String get name;
+
+  Expression get definition;
+
+  @override
+  R accept<R>(
+    SpecVisitor<R> visitor, [
+    R? context,
+  ]) =>
+      visitor.visitTypeDef(this, context);
+}
+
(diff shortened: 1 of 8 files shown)


lib/src/specs/typedef.dart
```diff
@@ -0,0 +1,55 @@
...
+  TypeDefBuilder._();
+
+  @override
+  ListBuilder<Expression> annotations = ListBuilder<Expression>();
+
+  @override
+  ListBuilder<String> docs = ListBuilder<String>();
+
+  /// Name of the field.
+  String? name;
```
natebosch, line 52:
```suggestion
  String? name;
```

lib/src/specs/typedef.dart
```diff
@@ -0,0 +1,55 @@
...
+    with HasAnnotations, HasDartDocs
+    implements Built<TypeDef, TypeDefBuilder>, Spec {
+  factory TypeDef([void Function(TypeDefBuilder)? updates]) = _$TypeDef;
+
+  TypeDef._();
+
+  /// Name of the typedef.
+  String get name;
+
+  Expression get definition;
```
natebosch, line 28:
```suggestion

  /// The right hand side of the typedef.
  ///
  /// Typically a reference to a type, or a Function type.
  Expression get definition;
```

natebosch (commented):
We probably need a way to add these on a library.

lib/src/specs/typedef.dart
```diff
@@ -0,0 +1,55 @@
...
+import '../visitors.dart';
+import 'expression.dart';
+
+part 'typedef.g.dart';
+
+@immutable
+abstract class TypeDef extends Object
+    with HasAnnotations, HasDartDocs
+    implements Built<TypeDef, TypeDefBuilder>, Spec {
+  factory TypeDef([void Function(TypeDefBuilder)? updates]) = _$TypeDef;
```
srawlins, line 21: It'd be great if this could support generic typdefs (`typedef Fn<T extends Foo> = T Function();`), but I understand maybe no one needs it yet.

I only bring it up as a consideration because it might be a breaking change to add support for generics, which is a painful upgrade to make to code_builder, especially when migrating clients and everything that depends on code_builder internally...

srawlins (approved):

lib/src/specs/typedef.dart
devoncarew: I'm a little less familiar with this library, but that adding generics support may just be adding `BuiltList<Reference> get types` to `TypeDef`? I think it would just parameterize the `name` field, and people could manually parameterize the `Expression? definition` on the RHS.

devoncarew:
> We probably need a way to add these on a library.

Yeah, I was using this locally, but I have a few different changes to the library locally - I may have missed some relevant pieces when uploading these changes for review.

I'll check locally and circle back to this PR.

devoncarew:
> I'll check on disk and circle back to this PR.

Ah, I was just using `library.body.add(typedef)` - which takes a `Spec` - the same way I was adding classes and such.

PR #389, opened 2022-12-16
add support for ignore_for_file comments

devoncarew:
- add support for emitting leading line comments for libraries
- add support for emitting `ignore_for_file` analyzer directive comments


lib/src/emitter.dart
```diff
@@ -398,6 +398,24 @@ class DartEmitter extends Object
   @override
   StringSink visitLibrary(Library spec, [StringSink? output]) {
     output ??= StringBuffer();
+
+    spec.comments.map((line) => '// $line').forEach(output.writeln);
+    output.writeln();
```
natebosch, line 403: [nit] Do we want an extra blank line if there were no comments?

natebosch (approved):

lib/src/emitter.dart
devoncarew: nope - I'll guard this w/ `if (spec.comments.isNotEmpty)`.

Issue #315, opened 2021-03-28
AssignFinal to nullable type doesn't generate the nullable type


johanflint:
When using code_builder I noticed that if I use `assignFinal` with a nullable type, it doesn't output the question mark.

Reproduction recipe:

```dart
final statement = refer('foo')
    .assignFinal(
        'bar',
        TypeReference((b) => b
          ..symbol = 'String'
          ..isNullable = true))
    .statement;
print(DartFormatter().format('${statement.accept(DartEmitter.scoped(useNullSafetySyntax: true))}'));
```

Output: `final String bar = foo;`
Expected: `final String? bar = foo;`

Is this a bug or am I missing something obvious here? Thanks for the great package, it's a joy to use!

srawlins:
Yeah I see a few issues with `assignFinal`, `assignConst`, and `assignVar`. They all use `TypeReference.expression`, which uses ScopedCode, which only references the `symbol` in its `toString`, not the type arguments, nor the nullability. So this:

```dart
void main() {
  final statement2 = refer('foo')
      .assignFinal(
          'bar',
          TypeReference((b) => b
            ..symbol = 'List'
            ..types.add(refer('int'))))
      .statement;
  print(DartFormatter().format(
      '${statement2.accept(DartEmitter.scoped())}'));
}
```

prints `final List bar = foo;`, omitting the `<int>` type argument.

johanflint:
Ah indeed. I use `...symbol = 'List?'` as a workaround, but it feels like cheating :).

Lohann:
still waiting for a fix

natebosch:
`assignFinal` and the related APIs are deprecated. It looks like this bug might exist in the replacements too

PR #401, opened 2023-03-07
Failing test for assigning to nullable typed var

natebosch:
Towards #315

Add a skipped test which demonstrates that the nullability of a type
reference is ignored when emitting a variable assignment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 4.4.1-dev
+
 ## 4.4.0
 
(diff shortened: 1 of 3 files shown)


kevmoo (approved):
Uh...adding a skipped test doesn't do...a lot.

natebosch:
> Too bad we can't pass a `failing: true` argument. But this works.

Yeah, it's been discussed a few times. Most recently at https://github.com/dart-lang/test/issues/1832


> adding a skipped test doesn't do...a lot.

It avoids the small amount of work already done to confirm that the bug exists in the new APIs.

PR #414, opened 2023-05-14
Class modifiers

dnys1:
An attempt at adding class modifiers in a non-breaking way.

Outstanding questions:
- Is the validation and/or API enough to encourage proper use?
- Are the tests sufficient for the changes?

---

#### Contribution guidelines:

Many Dart repos have a weekly cadence for reviewing PRs - please allow for a week or two of latency for initial review feedback.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.5.1-dev
+
+* Add support for class modifiers.
+
 ## 4.5.0
 
 * Require Dart 2.19
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 4.5.1-dev
+## 4.6.0-dev
 
 * Add support for class modifiers.
 
--- a/lib/src/specs/class.dart
+++ b/lib/src/specs/class.dart
@@ -30,6 +30,15 @@ abstract class Class extends Object
   /// Whether the class is `abstract`.
   bool get abstract;
 
+  /// Whether the class is `sealed`.
+  bool get sealed;
+
+  /// Whether the class is a `mixin class`.
+  bool get mixin;
+
+  /// The class modifier, i.e. `base`, `final`, `interface`.
+  ClassModifier? get modifier;
+
   @override
   BuiltList<Expression> get annotations;
 
@@ -60,6 +69,23 @@ abstract class Class extends Object
       visitor.visitClass(this, context);
 }
 
+enum ClassModifier {
+  base,
+  final$,
+  interface;
+
+  String get name {
+    switch (this) {
+      case ClassModifier.base:
+        return 'base';
+      case ClassModifier.final$:
+        return 'final';
+      case ClassModifier.interface:
+        return 'interface';
+    }
+  }
+}
+
 abstract class ClassBuilder extends Object
     with HasAnnotationsBuilder, HasDartDocsBuilder, HasGenericsBuilder
     implements Builder<Class, ClassBuilder> {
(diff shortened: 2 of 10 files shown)


lib/src/specs/class.dart
```diff
@@ -75,6 +101,15 @@ abstract class ClassBuilder extends Object
   /// Whether the class is `abstract`.
   bool abstract = false;
 
+  /// Whether the class is `sealed`.
+  bool sealed = false;
```
natebosch, line 105: Is there a reason that `sealed` isn't modeled as a `ClassModifier`?

natebosch: cc @srawlins - I'm curious if you have any thoughts on how to model class modifiers - is it worth trying to encode _some_ of the cross usage constraints by having some be an `enum`, or do you think we should model each modifier as a boolean?

lib/src/specs/class.dart
```diff
@@ -60,6 +69,23 @@ abstract class Class extends Object
       visitor.visitClass(this, context);
 }
 
+enum ClassModifier {
+  base,
+  final$,
+  interface;
+
+  String get name {
```
srawlins, line 77: Dunno if this is Dart 3.0 code yet, but you can use a fancy switch expression as well:

```dart
String get name =>
    return switch (this) {
      base => 'base',
      final$ => 'final',
      interface => 'interface',
    };

File: lib/src/specs/class.dart (hunk 1 above)
Comment (srawlins) [reply lib/src/specs/class.dart:105]: How this is implemented now works out pretty nicely, looking at [this table from the spec](https://github.com/dart-lang/language/blob/main/accepted/future-releases/class-modifiers/feature-specification.md#syntax).

This impl does encode a bit what is legal and what is not.

Building from scratch, it would probably be simplest to have one enum for all of the modifiers, and then store a List (or Set) of them. But that would not be backwards compatible?

I don't really have a strong opinion.

**Review by srawlins — APPROVED** [review]

**Review by srawlins — APPROVED** [review]

File: lib/src/specs/class.dart (hunk 2 above)
Comment (dnys1) [reply lib/src/specs/class.dart:77]: Good catch! I did try bumping the constraints to ^3.0.0 but it will require a larger effort as the repo depends on some of Dart 2 behavior like mixing in abstract classes.

File: lib/src/specs/class.dart [hunk 3]
```diff
@@ -60,6 +69,23 @@ abstract class Class extends Object
       visitor.visitClass(this, context);
 }
 
+enum ClassModifier {
+  base,
+  final$,
```
natebosch, line 74: @munificent - I'm curious how you would name something referring to the `final` keyword.

This would be the first place we use a trailing `$` to avoid an otherwise unusable word. We have used `clazz` in place of `class` in a few variables and an argument, but that is mostly not visible in user code.

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: code_builder
-version: 4.5.0
+version: 4.5.1-dev
```
natebosch, line 2:
```suggestion
version: 4.6.0-dev
```

lib/src/specs/class.dart
dnys1: This was modeled after `FieldModifier` but I would also be curious to hear @munificent's take!

https://github.com/dart-lang/code_builder/blob/e8c01b00a7327743cc01deaaba0e23346eced78d/lib/src/specs/field.dart#L61

pubspec.yaml
dnys1: Updated 👍

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 4.5.1-dev
```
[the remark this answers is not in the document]
dnys1: Updated 👍

lib/src/specs/class.dart
natebosch: Ah, I forgot about that one and I was failing at writing a regex to look for it. Given the precedent we should stick with this name.

natebosch (approved):

lib/src/specs/class.dart
munificent: Follow whatever convention makes sense. Personally, I tend to follow Analyzer and would do `finalKeyword`. It's a little verbose, but I like that more than arbitrarily mangling the name (`clas`, `clazz`, etc.).

PR #417, opened 2023-05-24
Records

### LINKED ISSUE #416: Records are not supported
AFAICT, there is no way to produce neither record types, nor record expressions, with the only option being the `CodeExpression` escape hatch. I'm going to try implementing it, filing the bug for the reference.

yanok:
- Added support for records feature in Dart 3.0. Both record types and literals are supported. Fixes #416.

---

#### Contribution guidelines:

(note that many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback)

### DIFF
--- a/lib/src/emitter.dart
+++ b/lib/src/emitter.dart
@@ -17,6 +17,7 @@ import 'specs/method.dart';
 import 'specs/mixin.dart';
 import 'specs/reference.dart';
 import 'specs/type_function.dart';
+import 'specs/type_record.dart';
 import 'specs/type_reference.dart';
 import 'specs/typedef.dart';
 import 'visitors.dart';
@@ -529,6 +530,36 @@ class DartEmitter extends Object
       out.write('}');
     }
     out.write(')');
+    if (_useNullSafetySyntax && (spec.isNullable ?? false)) {
+      out.write('?');
+    }
+    return out;
+  }
+
+  @override
+  StringSink visitRecordType(RecordType spec, [StringSink? output]) {
+    final out = (output ??= StringBuffer())..write('(');
+    visitAll<Reference>(spec.positionalFieldTypes, out, (spec) {
+      spec.accept(this, out);
+    });
+    if (spec.namedFieldTypes.isNotEmpty) {
+      if (spec.positionalFieldTypes.isNotEmpty) {
+        out.write(', ');
+      }
+      out.write('{');
+      visitAll<MapEntry<String, Reference>>(spec.namedFieldTypes.entries, out,
+          (entry) {
+        entry.value.accept(this, out);
+        out.write(' ${entry.key}');
+      });
+      out.write('}');
+    } else if (spec.positionalFieldTypes.length == 1) {
+      out.write(',');
+    }
+    out.write(')');
+    // It doesn't really make sense to use records without
+    // `_useNullSafetySyntax`, but since code_builder is generally very
+    // permissive, follow it here too.
     if (_useNullSafetySyntax && (spec.isNullable ?? false)) {
       out.write('?');
     }
--- a/lib/src/specs/type_record.dart
+++ b/lib/src/specs/type_record.dart
@@ -1,0 +1,99 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:built_collection/built_collection.dart';
+import 'package:built_value/built_value.dart';
+import 'package:meta/meta.dart';
+
+import '../base.dart';
+import '../visitors.dart';
+import 'expression.dart';
+import 'reference.dart';
+
+part 'type_record.g.dart';
+
+@immutable
+abstract class RecordType extends Expression
+    implements Built<RecordType, RecordTypeBuilder>, Reference, Spec {
+  factory RecordType([
+    void Function(RecordTypeBuilder) updates,
+  ]) = _$RecordType;
+
+  RecordType._();
+
+  @override
+  R accept<R>(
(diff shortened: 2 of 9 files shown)


lib/src/emitter.dart
```diff
@@ -535,6 +536,38 @@ class DartEmitter extends Object
...
+      spec.accept(this, out);
+    });
+    if (spec.namedFieldTypes.isNotEmpty) {
+      if (spec.positionalFieldTypes.isNotEmpty) {
+        out.write(', ');
+      }
+      out.write('{');
+      visitAll<MapEntry<String, Reference>>(spec.namedFieldTypes.entries, out,
+          (entry) {
+        out.write('${entry.key} ');
```
srawlins, line 552: These are backwards. See the illegal syntax generated in `record_type_test.dart`: `({named int, other List<int>})`

test/specs/code/expression_test.dart
```diff
@@ -141,6 +142,31 @@ void main() {
...
+          1,
+          refer('one'),
+          'one'
+        ], {
+          'named': refer('Foo').newInstance([literalNum(1)])
+        }),
+        equalsDart("(1, one, 'one', named: Foo(1))"));
+  });
+
+  test('should correctly emit a record with a single positional field', () {
```
srawlins, line 166: Great tests, can I ask for two more, just because of the delicate commas:

* empty record, which is allowed (`()`)
* record with only named fields

lib/src/emitter.dart
yanok: Good catch! That's how you write bad code _and_ bad tests ;) Fixed.

test/specs/record_type_test.dart
```diff
@@ -0,0 +1,65 @@
...
+        ..positionalFieldTypes.addAll(
+            [intRef, listRef(intRef).rebuild((b) => b..isNullable = true)])
+        ..isNullable = true),
+      equalsDart('(int, List<int>?)?', emitter),
+    );
+  });
+
+  test('should create a record type with one positional field', () {
+    expect(RecordType((b) => b..positionalFieldTypes.add(intRef)),
+        equalsDart('(int,)', emitter));
```
srawlins, line 35: These are great, can we add one for the empty record type (`()`) (Just looked it up, very strange but it's legal)

srawlins (commented):
LGTM, very idiomatic I love it. But I'll let @natebosch review as well.

test/specs/code/expression_test.dart
yanok: Sure! I added an empty record. There already was a test for only named fields, but I added/renamed some tests to make it clearer that we test all three cases of only positional, only named and mixed.

kevmoo (commented):
Are we sure we have the right constraint on pkg:analyzer? Are we using APIs that were introduced in a version above the lowest constraint?

Just curious...

yanok:
> Are we sure we have the right constraint on pkg:analyzer? Are we using APIs that were introduced in a version above the lowest constraint?

but this package doesn't depend on analyzer...

lib/src/emitter.dart
```diff
@@ -535,6 +536,38 @@ class DartEmitter extends Object
...
+      }
+      out.write('{');
+      visitAll<MapEntry<String, Reference>>(spec.namedFieldTypes.entries, out,
+          (entry) {
+        entry.value.accept(this, out);
+        out.write(' ${entry.key}');
+      });
+      out.write('}');
+    } else {
+      if (spec.positionalFieldTypes.length == 1) {
```
natebosch, line 557: [nit] can we use `} else if (` to remove 1 level of indentation like you did in `expression.dart`

lib/src/specs/type_record.dart
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
natebosch, line 1: [nit] Why `type_record.dart` instead of `record_type.dart`?

lib/src/specs/type_record.dart
```diff
@@ -0,0 +1,104 @@
...
+  RecordType._();
+
+  @override
+  R accept<R>(
+    SpecVisitor<R> visitor, [
+    R? context,
+  ]) =>
+      visitor.visitRecordType(this, context);
+
+  /// Positional fields types.
```
natebosch, line 32: [nit] remove or expand to include details not already in the name

lib/src/specs/type_record.dart
```diff
@@ -0,0 +1,104 @@
...
+  BuiltList<Reference> get positionalFieldTypes;
+
+  /// Named fields types.
```
natebosch, line 35: [nit] remove or expand to include details not already in the name

lib/src/specs/type_record.dart
```diff
@@ -0,0 +1,104 @@
...
+
+  @override
+  String? get symbol => null;
+
+  @override
+  Reference get type => this;
+
+  /// Optional nullability.
+  ///
+  /// An emitter may ignore this if the output is not targeting a Dart language
```
natebosch, line 49: [nit] I would drop this detail. It's fine to include the handling when emitting, but I wouldn't introduce this as a concept in the docs since it isn't sensible.

natebosch (approved):

lib/src/specs/type_record.dart
yanok: I was following the `type_function.dart` example, but I don't have a strong opinion here.

lib/src/specs/type_record.dart
yanok: Done.

lib/src/specs/type_record.dart
yanok: Dropped.

Issue #409, opened 2023-03-24
Add external key word for the FieldBuilder


XinyuKhan:
Something like this:

```Dart
class MyStruct externs Struct {
    @Float()
    external double value;
}
```

PR #422, opened 2023-06-21
added support of external keyword for field builder

Snehal-Singh174:
- Added support of external keyword for FieldBuilder
-  Issue #409 
- Added test case for external keyword inside field_test
- ```test('should create a external field', () {
    expect(
      Field((b) => b
        ..name = 'value'
        ..external = true
        ..type = refer('double')
        ..annotations.addAll([
          refer('Float').call([])
        ])),
      equalsDart(r'''
        @Float()
        external double value;
      '''),
    );
  });

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.7.0
+
+* Add support of external keyword for fieldBuilder
+
 ## 4.6.0-wip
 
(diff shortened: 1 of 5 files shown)


natebosch:
I think the addition should be OK. Can you format with `dart format` to fix the CI error?

We will also need a CHANGELOG entry.

natebosch (approved):

PR #425, opened 2023-08-04
Add Expression.parenthesized

natebosch:
We use `ParenthesizedExpression` internally in a few places where it's
know that they are required, but some corners of the syntax require
extra parenthesis in ways that are not feasible to detect in this
library. Allow for explicit manual parenthesis.


srawlins (approved):
I've experienced this; this'll be hlpful!

PR #433, opened 2023-10-04
Allow passing an `Expression` through `literal`

natebosch:
The `LiteralListExpression` allows a mix of Dart values which could be
passed to `literal` and values which are already a `Spec` (typically an
`Expression`). Both types are handled by `_acceptLiteral` in the visitor
implementation.

https://github.com/dart-lang/code_builder/blob/eb70874cf05427879ec47d8726eb3173e11880e3/lib/src/specs/expression.dart#L622-L628

The `toString()` for `LiteralListExpression` calls `literal` on each
value in the list, which previously would throw if it was passed an
`Expression`. Some tests for builders and their implementation details
use `.toString()` in their user facing failure messages. When there is
an exception creating the message for a failure, it can mask the cause
of the failure.

Allow more flexibility for this `toString` by allowing `Expression`
instances to flow through `literal` unchanged. This still does not
handle non-expression `Spec` instances, but those are not likely to come
up in practice.

Collapse the conditional returns in `literal` to a single line each. The
pattern is easier to read with less vertical space.

Add tests for `literal` passing each type it accepts, and a test for the
error behavior.

Add a regression test for the `toString` of a list literal not throwing
when one of the values of the list is an `Expression`.

### DIFF
--- a/lib/src/specs/expression/literal.dart
+++ b/lib/src/specs/expression/literal.dart
@@ -8,30 +8,15 @@ part of '../expression.dart';
 ///
 /// Unsupported inputs invoke the [onError] callback.
 Expression literal(Object? literal, {Expression Function(Object)? onError}) {
-  if (literal is bool) {
-    return literalBool(literal);
-  }
-  if (literal is num) {
-    return literalNum(literal);
-  }
-  if (literal is String) {
-    return literalString(literal);
-  }
-  if (literal is List) {
-    return literalList(literal);
-  }
-  if (literal is Set) {
-    return literalSet(literal);
-  }
-  if (literal is Map) {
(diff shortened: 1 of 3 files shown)


lib/src/specs/expression/literal.dart
```diff
@@ -8,30 +8,15 @@ part of '../expression.dart';
 ///
```
jakemac53, line 8: I think its probably worth adding a comment that Expression is also handled and why

natebosch: Done.

jakemac53 (approved):

srawlins (commented):
❤️

PR #437, opened 2023-11-17
Add support for extension types

srujzs:
Corresponds to the syntax defined in https://github.com/dart-lang/language/blob/main/accepted/future-releases/extension-types/feature-specification.md#syntax.

- [X] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

### DIFF
--- a/lib/src/emitter.dart
+++ b/lib/src/emitter.dart
@@ -11,6 +11,7 @@ import 'specs/directive.dart';
 import 'specs/enum.dart';
 import 'specs/expression.dart';
 import 'specs/extension.dart';
+import 'specs/extension_type.dart';
 import 'specs/field.dart';
 import 'specs/library.dart';
 import 'specs/method.dart';
@@ -334,6 +335,62 @@ class DartEmitter extends Object
     }
     out.writeln(' }');
     return out;
+  }
+
+  @override
+  StringSink visitExtensionType(ExtensionType spec, [StringSink? output]) {
+    final out = output ??= StringBuffer();
+    spec.docs.forEach(out.writeln);
+    for (var a in spec.annotations) {
+      visitAnnotation(a, out);
+    }
+
+    out.write('extension type ');
+    if (spec.constant) out.write('const ');
+    out.write(spec.name);
+    visitTypeParameters(spec.types.map((r) => r.type), out);
+    if (spec.primaryConstructorName.isNotEmpty) {
+      out.write('.${spec.primaryConstructorName}');
+    }
+    out.write('(');
+    _visitRepresentationDeclaration(spec.representationDeclaration, out);
+    out.write(')');
+
+    if (spec.implements.isNotEmpty) {
+      out
+        ..write(' implements ')
+        ..writeAll(
+            spec.implements.map<StringSink>((m) => m.type.accept(this)), ',');
+    }
+
+    out.writeln(' {');
+    for (var c in spec.constructors) {
+      visitConstructor(c, spec.name, out);
+      out.writeln();
+    }
+    for (var f in spec.fields) {
+      visitField(f, out);
+      out.writeln();
+    }
+    for (var m in spec.methods) {
+      visitMethod(m, out);
+      if (_isLambdaMethod(m)) {
+        out.writeln(';');
+      }
+      out.writeln();
+    }
+    out.writeln('}');
+    return out;
+  }
+
+  void _visitRepresentationDeclaration(
+      RepresentationDeclaration spec, StringSink out) {
+    spec.docs.forEach(out.writeln);
+    for (var a in spec.annotations) {
+      visitAnnotation(a, out);
+    }
+    spec.declaredRepresentationType.accept(this, out);
+    out.write(' ${spec.name}');
   }
 
   @override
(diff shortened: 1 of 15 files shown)


test/specs/extension_test.dart
```diff
@@ -134,25 +134,4 @@ void main() {
       '''),
     );
   });
-
-  test('should create an extension with a method', () {
```
srujzs, line 138: FYI - this is just a duplicate test.

pubspec.yaml
```diff
@@ -19,6 +19,11 @@ dev_dependencies:
   build_runner: ^2.0.3
   built_value_generator: ^8.0.0
   dart_flutter_team_lints: ^1.0.0
-  dart_style: ^2.0.0
+  # TODO(srujzs): Restore this to ^2.0.0 once 2.3.4 is released and format the
```
srujzs, line 22: I'm not really sure if this is an issue (and it looks like CI isn't happy with it anyways due to the SDK requirement). Should I just keep it as `^2.0.0`?

lib/src/emitter.dart
```diff
@@ -336,6 +337,62 @@ class DartEmitter extends Object
...
+    }
+    out.write('(');
+    _visitRepresentationDeclaration(spec.representationDeclaration, out);
+    out.write(')');
+
+    if (spec.implements.isNotEmpty) {
+      out
+        ..write(' implements ')
+        ..writeAll(
+            spec.implements.map<StringSink>((m) => m.type.accept(this)), ', ');
```
jakemac53, line 363: consider joining with ', ' instead so we don't have a trailing comma?

srujzs: It is indeed what it does - it takes in a separator between any two elements. I do add a space here to make the non-formattable code look a bit better, but maybe I shouldn't for consistency and just wait for the new version of dart_style.

jakemac53:
@srujzs fwiw force pushing in GitHub doesn't work well because it destroys all history (I can't tell what you actually changed in that last push). If you can do a merge from master instead in the future that will help me out :).

pubspec.yaml
jakemac53: Yeah just keep it at ^2.0.0 (it will still get the latest that it can, on whatever SDK it is running).

We could update the SDK versions of the the job is running the formatter if that is desirable, but the important thing is that your formatter version matches that of the job we are running, otherwise you will have issues.

test/specs/extension_type_test.dart
```diff
@@ -0,0 +1,242 @@
...
+        ..name = 'Foo'
+        ..implements.addAll([
+          TypeReference((b) => b.symbol = 'num'),
+          TypeReference((b) => b.symbol = 'Object')
+        ])
+        ..representationDeclaration = RepresentationDeclaration((b) => b
+          ..declaredRepresentationType = TypeReference((b) => b.symbol = 'int')
+          ..name = 'bar')),
+      equalsDart(r'''
+        extension type Foo(int bar) implements num,Object { }
```
jakemac53, line 162: I would expect a space between these, the matcher might just be a bit loose though because I thought you did add that, but might as well update the expectation to show it:

```suggestion
        extension type Foo(int bar) implements num, Object { }
```

jakemac53 (commented):
LGTM modulo the formatter change

pubspec.yaml
jakemac53: Oh I should have read this comment more fully :).

We should probably increase the minimum to require a version of dart_style that can actually format extension types.

I just chatted with @munificent and it sounds like we can get a release, so we could change this to ^2.3.4.

srujzs:
> fwiw force pushing in GitHub doesn't work well because it destroys all history (I can't tell what you actually changed in that last push).

Ah sorry about that! I've been pushing to avoid unnecessary commits from polluting the branch but messing with the review is way worse.

test/specs/extension_type_test.dart
srujzs: This was the one change I made before the force push. :)

I changed the separator to not have that space to be consistent with the rest of the emitter and to avoid spaces that aren't needed in the output. Dart format should add the space for us when 2.3.4 is released.

pubspec.yaml
srujzs: Got it. I guess wait for 2.3.4 then before landing this? I'll do some testing with 2.3.3 and updated SDK constraint to see if any of the bots are unhappy first.

jakemac53:
> Ah sorry about that! I've been pushing to avoid unnecessary commits from polluting the branch but messing with the review is way worse.

No worries, really it is just specific to GitHub and how its reviews work. Fwiw when we merge we squash/merge so its always just one commit. So you don't have to worry about branch history really (and it is useful for reviewing so you can see diffs since last review).

test/specs/extension_type_test.dart
jakemac53: Ah ok, SGTM :)

jakemac53 (approved):

pubspec.yaml
jakemac53: as discussed offline either way is OK with me :)

PR #439, opened 2023-11-29
support empty line comments

devoncarew:
- support empty line comments

This is to allow us to do things like:

```
// Copyright foo
// bar
// baz.

// Here's an additional file comment, but separated from the copyright header.
```

---

<details>
  <summary>Contribution guidelines:</summary><br>

Note that many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.
</details>


test/specs/library_test.dart
```diff
@@ -45,7 +45,7 @@ void main() {
         ),
         equalsDart(r'''
           // Generated by foo!
-          //
```
srawlins, line 48: It seems like there should still be a way to get this output. Not sure how though... `comments` could be a list of nullable strings, and then null means something different from a blank string. Or a string that is only whitespace vs blank? I can't think of anything great...

devoncarew: I don't know that it's an important use case to support `//` lines? But the easiest and most clear way would likely be by supporting null strings.

An alternative approach would be to support line wrapping. So, multiple lines would mean multiple groups of comments, with each line automatically wrapped at ~80 cols.

I may implement the 1st solution as that seems like the smallest API diff.

devoncarew:
ptal - updated to support both empty line comments as well as having groups of line comments

test/specs/library_test.dart
srawlins: Well, I don't have any examples, but I can imagine that it's more common to write a multi-paragraph comment (with blank `//` lines) than adjacent comments. Like if you have two paragraphs, or an example, like a code block separated from test.

srawlins:
I'm not a huge fan of the nullable String API, where null is sort of a "magic value." But I do think it is extremely pragmatic, functional, easy to understand...

I think I'd like another opinion like @natebosch before landing. If we go a nullable String API then it would be a pain to bring it back to a non-nullable String API. I'd almost prefer an API where "a comment" is a `List<String>` and a declaration, like a class, can have "plural comments", so a `List<Comment>`. But that might not be pragmatic, making everyone write so many nested structures just to add a little one-line comment somewhere...

devoncarew:
> I'm not a huge fan of the nullable String API, where null is sort of a "magic value." But I do think it is extremely pragmatic, functional, easy to understand...

Agree w/ the nullability aspect; I think it doesn't improve the API. I've reverted back to a simplier form of this PR; happy to wait for an opinion from Nate.

For context, this PR is to improve the output of the files we generate for package:web.

devoncarew:
Thinking about the comments above:

- I don't love changing the meaning of the current comments field (what this PR proposes and in Nate's idea above)
- I think a green field solution would use something like the `List<Comment>` idea from above, but I don't think the new api / deprecated api dance for a new `comments` field is worth it for this feature
- the use case I have is wanting both a file copyright section and a 'generated by' line for a generated file (and for that 2nd comment to be separate from the copyright one)

I have a separate PR now - https://github.com/dart-lang/code_builder/pull/441 - which adds a new field for that specific use case - a `generatedByComment` field. This is not dissimilar to the existing `Library.ignoreForFile` field.

PR #441, opened 2023-12-11
Add a new field `Library.generatedByComment` to support emitting 'generated by' comments.

devoncarew:
- Add a new field `Library.generatedByComment` to support emitting 'generated by' comments.

This is an alternative to https://github.com/dart-lang/code_builder/pull/439.

---

<details>
  <summary>Contribution guidelines:</summary><br>

Note that many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.
</details>


srawlins (approved):
Oh hey, I really like this idea. This sounds really slick to me.

lib/src/specs/library.dart
```diff
@@ -30,6 +30,9 @@ abstract class Library
   /// Line comments to place at the start of the library.
   BuiltList<String> get comments;
 
+  /// A comment indicating the tool this library was generated by.
```
natebosch, line 33: Can we consider adding even more convention, and asking only for the tool name and then fill in the "Generated by" part ourselves?

Either way, I think we should have an example in the doc comment so folks know how much of the string they need to provide (currently, they need to provide the "Generated by" words in addition to the tool name, but they don't provide the `//`)

devoncarew: I updated the dartdoc comments for clarity about what content to put in the message.

PR #463, opened 2024-06-06
Support newInstanceNamed with empty name

mattrberry:
Opts to rewrite  `refer('Foo').newInstanceNamed('', [])` as `Foo` rather than `Foo.`. Two primary reasons
1. `Foo.` breaks compilation with bad error messages
2. [Analyzer's `ConstructorElement#name`](https://github.com/dart-lang/sdk/blob/e8939b0ff089a97297cf28626be46c3c75fe72e7/pkg/analyzer/lib/dart/element/element.dart#L466) is a non-nullable String, which pushes the `isEmpty` check to user code

---

### DIFF
--- a/lib/src/specs/expression.dart
+++ b/lib/src/specs/expression.dart
@@ -578,7 +578,7 @@ abstract mixin class ExpressionEmitter
     final out = output ??= StringBuffer();
     return _writeConstExpression(out, expression.isConst, () {
       expression.target.accept(this, out);
-      if (expression.name != null) {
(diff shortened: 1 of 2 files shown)


srawlins (approved):
Thanks! ❤️ 

I'm not a primary owner so I think we should wait for one more review to land.

mosuem:
Closing as the [dart-lang/code_builder](https://github.com/dart-lang/code_builder) repository is merged into the [dart-lang/tools](https://github.com/dart-lang/tools) monorepo. Please re-open this PR there!

PR #466, opened 2024-10-22
Update dart_style to pass in a language version to DartFormatter.

munificent:
We're in [the process of](https://github.com/dart-lang/dart_style/issues/1403) moving dart_style to [a new formatting style](https://github.com/dart-lang/dart_style/issues/1253). That involves making the formatter [aware of the language version of what it's formatting](https://github.com/dart-lang/dart_style/issues/1402).

That in turn means that the library API now lets you pass in a language version. In dart_style 2.3.7 [you can pass in a language version but the parameter is optional](https://pub.dev/documentation/dart_style/latest/dart_style/DartFormatter/DartFormatter.html). In the forthcoming 3.0.0 release, that parameter will become mandatory.

This updates every call to `DartFormatter()` to pass in the latest language version. If there's a more specific version that should be used, let me know and I'll update the PR.

Thanks!

### DIFF
--- a/test/common.dart
+++ b/test/common.dart
@@ -5,12 +5,15 @@
 import 'package:code_builder/code_builder.dart';
 import 'package:dart_style/dart_style.dart';
 
-final DartFormatter _dartfmt = DartFormatter();
 String _format(String source) {
+  final formatter = DartFormatter(
+    languageVersion: DartFormatter.latestLanguageVersion,
+  );
+
   try {
-    return _dartfmt.format(source);
+    return formatter.format(source);
   } on FormatterException catch (_) {
-    return _dartfmt.formatStatement(source);
+    return formatter.formatStatement(source);
   }
 }
 
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -43,7 +43,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        sdk: [3.3.0, dev]
+        sdk: [3.5.0, dev]
     steps:
       - uses: actions/checkout@d632683dd7b4114ad314bca15554477dd762a938
       - uses: dart-lang/setup-dart@0a8a0fc875eb934c15d08629302413c671d3f672
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -14,7 +14,7 @@ env:
 
 jobs:
   # Check code formatting and static analysis on a single OS (linux)
-  # against Dart dev and 2.12.0.
+  # against Dart dev and an earlier stable version.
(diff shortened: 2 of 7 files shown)


test/common.dart
```diff
@@ -5,12 +5,15 @@
 import 'package:code_builder/code_builder.dart';
 import 'package:dart_style/dart_style.dart';
 
-final DartFormatter _dartfmt = DartFormatter();
```
srujzs, line 8: It's hard to discern the use cases of this, but it may make sense to keep this as a top-level so we're not reinstantiating every time we call `_format`.

srujzs (approved):

test/common.dart
munificent: I admit this is maybe not the best API design, but `DartFormatter` has always been a very lightweight, ephemeral class. It's basically so that you don't have to pass a big pile of arguments to `format()`. It's fine to construct a new one on every invocation. And, arguably, it's better to do so because it ensures that the tiny amount of mutable state inside the class isn't shared across unrelated formats (https://github.com/dart-lang/dart_style/issues/1337).

devoncarew (approved):
Can you add in a line to the latest changelog entry? Something like `Upgrade to dart_style 2.3.7.`.

Also, it looks like there is an existing analysis issue w/ the analysis_options.yaml file; it would be great to address that as well, so that we can land this PR on green.

From your PR description and the code change, it looks like we're fine w/o this change as the upcoming api break is in the next major version of dart_style; but perhaps this makes it easier to roll that new version into a monorepo w/ this package.

test/common.dart
srujzs: Ack, this is fine then. Thanks!

munificent:
> Can you add in a line to the latest changelog entry? Something like `Upgrade to dart_style 2.3.7.`.

Done.

> Also, it looks like there is an existing analysis issue w/ the analysis_options.yaml file; it would be great to address that as well, so that we can land this PR on green.

Sure thing, done.

> From your PR description and the code change, it looks like we're fine w/o this change as the upcoming api break is in the next major version of dart_style;

Yeah, you don't *need* to pass the language version in just to please 2.3.7, but you will to work with 3.0.0 which is hot on its heels.

> but perhaps this makes it easier to roll that new version into a monorepo w/ this package.

Yes, that's the motivation for me fixing this. :)

munificent:
@devoncarew, in order to get CI happy, I had to bump the stable version being tested to 3.3.0. Are you OK with that?

It should be fine since, after this change, anyone using the latest code_builder will need to be on 3.3.0 or later anyway (since it appears there is no version of analyzer that dart_style 2.3.7 accepts which allows an older SDK.

